PR #12, opened 2024-09-10
Glob import support

bioball:
This adds support for glob imports!

This PR is based on top of https://github.com/apple/pkl-lsp/pull/11, and should be reviewed after that other one gets merged.

### DIFF
--- a/src/main/kotlin/org/pkl/lsp/features/GoToDefinitionFeature.kt
+++ b/src/main/kotlin/org/pkl/lsp/features/GoToDefinitionFeature.kt
@@ -63,9 +63,9 @@ class GoToDefinitionFeature(private val server: PklLSPServer, project: Project) 
 
   private fun resolveModuleUri(
     originalNode: PklStringConstant,
+    parent: PklModuleUriOwner,
     context: PklProject?,
   ): List<LocationLink> {
-    val parent = originalNode.parent as PklModuleUriOwner
     val stringContentsSpan = originalNode.contentsSpan()
     if (parent is PklImportBase && parent.isGlob) {
       val resolved = parent.moduleUri?.resolveGlob(context) ?: return listOf()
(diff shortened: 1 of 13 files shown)


src/main/kotlin/org/pkl/lsp/services/PklProjectManager.kt
```diff
@@ -75,15 +75,15 @@ class PklProjectManager(project: Project) : Component(project) {
     if (file !is FsFile) null else pklProjects[file.projectKey]
 
   /** Tracks when PklProject files get added or deleted from the workspace. */
-  val addedOrRemovedFilesModificationTracker = SimpleModificationTracker()
+  val addedOrRemovedFilesModificationTracker =
+    project.pklFileTracker.filter { event -> event.files.any { it.path.endsWith("PklProject") } }
```
stackoverflow, line 79: This is not related to glob support, right?

stackoverflow (approved):
Just some small comments. Approving to unblock.

src/main/kotlin/org/pkl/lsp/features/GoToDefinitionFeature.kt
```diff
@@ -31,40 +32,78 @@ import org.pkl.lsp.type.computeThisType
...
     }
     return server.builder().runningBuild(params.textDocument.uri).thenApply(::run)
   }
 
-  private fun resolveDeclaration(
+  private fun resolveModuleUri(
+    originalNode: PklStringConstant,
+    context: PklProject?,
+  ): List<LocationLink> {
+    val parent = originalNode.parent as PklModuleUriOwner
```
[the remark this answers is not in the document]
bioball: Good point!

src/main/kotlin/org/pkl/lsp/services/PklProjectManager.kt
bioball: This is only related in that I needed to introduce another file tracker that just tracked when Pkl files get added or removed from the workspace. We already have other trackers for added/removed files, so I introduced a tracker just for adding/removing Pkl files, and made this one simply derived from it.

PR #13, opened 2024-09-11
Make LSP configurable using initialization options

bioball:
This defines initialization options that clients can send during the initialize request,
to configure the behavior of the server.

This also introduces extended capabilities that clients can notify the server about.

Also:

* Switch some messages from `sendActionableNotification` to `showMessageRequest`, which is a built-in message.
* Look for Pkl CLI on $PATH if not configured by client.
* Render doc comments as normal URIs instead of `command:pkl.open.file`, unless configured to do so.

### DIFF
--- a/src/main/kotlin/org/pkl/lsp/ast/Extensions.kt
+++ b/src/main/kotlin/org/pkl/lsp/ast/Extensions.kt
@@ -15,8 +15,12 @@
  */
 package org.pkl.lsp.ast
 
+import java.net.URI
 import java.net.URLEncoder
+import kotlinx.serialization.json.add
+import kotlinx.serialization.json.buildJsonArray
 import org.eclipse.lsp4j.Location
+import org.eclipse.lsp4j.LocationLink
 import org.pkl.lsp.*
 import org.pkl.lsp.FsFile
 import org.pkl.lsp.LSPUtil.toRange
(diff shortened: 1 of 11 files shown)


src/main/kotlin/org/pkl/lsp/ast/Extensions.kt
```diff
@@ -397,21 +399,36 @@ fun PklNode.findBySpan(line: Int, col: Int, includeTerminals: Boolean = false):
...
+  val useCommandLink = project.clientOptions.renderOpenFileCommandInDocs ?: false
+  val span = beginningSpan()
+  return when {
+    useCommandLink && forDocs -> {
+      val params = buildJsonArray {
+        add(containingFile.lspUri.toString())
+        add(span.beginLine)
+        add(span.beginCol)
+      }
+      URI("command", "pkl.open.file?$params", null)
```
stackoverflow, line 424: Have you checked this works without URLEncode? I also had problems if there was any spaces in the json, like after a `,`.

stackoverflow (approved):
LGTM

src/main/kotlin/org/pkl/lsp/ast/Extensions.kt
bioball: Yeah, the second argument here is the raw path. It gets encoded when you `toString()` the URI.

PR #14, opened 2024-09-13
Remove builder interface

bioball:
The goal of this PR is to separte the concerns of building a file from analysis and publishing diagnostics.

Each VirtualFile now returns the underlying PklModule as a completable future, which gets invalidated if its contents change.

* Add `getModule(): CompletableFuture<PklModule>` to VirtualFile
* Add `DiagnosticsManager`, which builds new diagnostics as the project's state changes
* Remove Stdlib interface (build stdlib module on demand, like any other module)
* Don't file parser diagnostic events, because I anticipate that this logic will get replaced once we've swapped out the ANTLR parser with tree-sitter.

Miscellaneous: change various events to follow ADT pattern


src/main/kotlin/org/pkl/lsp/Stdlib.kt
```diff
@@ -1,80 +0,0 @@
...
-package org.pkl.lsp
-
-import java.io.IOException
-import java.net.URI
-import org.pkl.core.parser.Parser
-import org.pkl.core.util.IoUtils
-import org.pkl.lsp.ast.PklModule
-import org.pkl.lsp.ast.PklModuleImpl
-
-class Stdlib(project: Project) : Component(project) {
```
bioball, line 25: I removed this because it's not really needed anymore, because every file is tracked by `VirtualFileManager` via its URI.

So, getting the base module is simply `virtualFileManager.get(URI("pkl:base")).get!!`, and that will return the same module every time.

This also means that we no longer need to initialize the standard library upon startup.

src/main/kotlin/org/pkl/lsp/VirtualFile.kt
```diff
@@ -75,19 +79,58 @@ interface VirtualFile : ModificationTracker {
...
+    set(text) {
+      myContents = text
+    }
+
+  final override fun getModule(): CompletableFuture<PklModule?> {
+    return project.cachedValuesManager.getCachedValue("VirtualFile($uri)") {
+      val future = CompletableFuture.supplyAsync { doBuildModule() }
+      CachedValue(future, this)
+    }!!
+  }
```
bioball, line 113: Using the cached values API with futures means that if the file is still valid (i.e. its contents have not changed), it will return the same future, which means we only build the module once.

The virtual file itself is the cache dependency; updating the virtual file will invalidate the cache, and a new future is computed.

src/main/kotlin/org/pkl/lsp/VirtualFileManager.kt
```diff
@@ -18,20 +18,9 @@ package org.pkl.lsp
...
-  override fun initialize(): CompletableFuture<*> {
-    project.messageBus.subscribe(textDocumentTopic) { event ->
-      val file = files[event.file] ?: return@subscribe
-      if (event.type == TextDocumentEventType.CHANGED) {
-        file.modificationCount.incrementAndGet()
-      }
-    }
-    return CompletableFuture.completedFuture(Unit)
-  }
-
```
bioball, line 34: I realized that the LSP tells us about the version of these files when they get opened; see https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocumentItem

So, I changed the TextDocumentService to just update `version` on each `VirtualFile`, and this version can be used as our modification count.

src/main/kotlin/org/pkl/lsp/ast/Extensions.kt
```diff
@@ -104,12 +104,8 @@ fun PklNode.isAncestor(of: PklNode): Boolean {
...
 }
 
-fun PklNode.isInStdlib(): Boolean {
-  return this.enclosingModule?.let { mod ->
-    val uri = mod.uri
-    uri.scheme == "pkl" && uri.authority == "stdlib"
-  } ?: false
-}
+val PklNode.isInStdlib
+  get(): Boolean = this.containingFile is StdlibFile
```
bioball, line 108: Changed this because vals with getters is more in line with Kotlin style (getters should be used for cheap computations)

src/main/kotlin/org/pkl/lsp/features/CompletionFeature.kt
```diff
@@ -132,7 +134,7 @@ class CompletionFeature(private val server: PklLSPServer, project: Project) : Co
     }
   }
 
-  private fun PklClass.complete(context: PklProject?): List<CompletionItem> = buildList {
+  private fun PklClass.complete(): List<CompletionItem> = buildList {
```
bioball, line 137: Unrelated: removed some parameters that weren't needed.

src/main/kotlin/org/pkl/lsp/util/FileCacheManager.kt
```diff
@@ -1,71 +0,0 @@
...
-      try {
-        val file = project.virtualFileManager.get(uri)!!
-        project.builder.fileToModule(contents, uri, file)
-      } catch (e: LexParseException) {
-        httpsErrors[uri] = errObject
-        return null
-      }
-    }
-  }
-}
```
bioball, line 71: I removed this because now that HTTPS files are tracked by VirtualFileManager, we don't need to write them to disk. Instead, we can cache the contents on the `VirtualFile` directly (see `doReadContents()`).

src/test/kotlin/org/pkl/lsp/LSPTestBase.kt
```diff
@@ -168,35 +165,14 @@ abstract class LSPTestBase {
...
-    val moduleCtx = parser.parseModule(contents)
-    return PklModuleImpl(moduleCtx, FsFile(path, fakeProject)).also { modules[path.toUri()] = it }
-  }
-
-  private fun parseAndStoreJar(contents: String, path: Path): PklModule {
-    val moduleCtx = parser.parseModule(contents)
-    val uri = path.toUri()
-    return PklModuleImpl(moduleCtx, JarFile(path, uri, fakeProject)).also { modules[uri] = it }
-  }
-
```
bioball, line 199: Since virtual files now handle their own parsing, we don't need this logic anymore.

src/main/kotlin/org/pkl/lsp/VirtualFile.kt
```diff
@@ -75,19 +79,58 @@ interface VirtualFile : ModificationTracker {
...
+  private val parser = Parser()
+
+  private fun doBuildModule(): PklModule? {
+    return try {
+      logger.log("building $uri")
+      val moduleCtx = parser.parseModule(contents)
+      return PklModuleImpl(moduleCtx, this)
+    } catch (e: LexParseException) {
+      logger.error("Parser Error building $file: ${e.message}")
+      null
```
bioball, line 128: Note: previously, this was building parser diagnostics and sending them back to the client.

I was considering keeping the same logic, but `LexParseException` is going away once we swap out the parser.

src/main/kotlin/org/pkl/lsp/PklTextDocumentService.kt
```diff
@@ -28,58 +28,57 @@ import org.pkl.lsp.services.Topic
...
-  CLOSED,
-  SAVED,
+  data class Opened(override val file: URI) : TextDocumentEvent
+
+  data class Closed(override val file: URI) : TextDocumentEvent
+
+  data class Saved(override val file: URI) : TextDocumentEvent
+
+  data class Changed(override val file: URI, val changes: List<TextDocumentContentChangeEvent>) :
+    TextDocumentEvent
```
bioball, line 41: I wanted to include more details based on the event type, so, using an ADT here is a lot nicer. I changed all the events to follow the ADT pattern.

src/main/kotlin/org/pkl/lsp/LSPUtil.kt
```diff
@@ -130,3 +130,16 @@ val URI.effectiveUri: URI?
...
+fun String.getIndex(position: Position): Int {
+  var currentIndex = 0
+  for ((column, line) in lines().withIndex()) {
+    if (column == position.line) {
+      return currentIndex + position.character
+    }
+    currentIndex += line.length + 1 // + 1 because newline is also a character
+  }
+  throw IllegalArgumentException("Invalid position for contents")
+}
```
bioball, line 145: Moved this method over from our tests, because we also need it in `PklTextDocumentService` to handle change events.

src/main/kotlin/org/pkl/lsp/PklLSPServer.kt
```diff
@@ -133,15 +120,6 @@ class PklLSPServer(val verbose: Boolean) : LanguageServer {
...
-    return project.pklProjectManager.syncProjects()
-  }
-
-  private fun cacheStdlib() {
-    stdlibDir.toFile().mkdirs()
-    for ((name, _) in project.stdlib.allModules()) {
-      val file = stdlibDir.resolve("$name.pkl")
-      val text = IoUtils.readClassPathResourceAsString(javaClass, "/org/pkl/core/stdlib/$name.pkl")
-      Files.writeString(file, text, Charsets.UTF_8)
-    }
```
bioball, line 145: Since the `VirtualFile`s themselves are now responsible for building, we don't really need to cache the stdlib anymore. Instead, we can just build them on demand, like we do any other module.

src/main/kotlin/org/pkl/lsp/Stdlib.kt
```diff
@@ -15,66 +15,28 @@
...
-      "settings",
-      "shell",
-      "test",
-      "xml",
-      "yaml",
-    )
-
-  init {
-    loadStdlib()
-  }
```
bioball, line 79: This is much simplified, and now works provides `VirtualFile` instead of `PklModule` (the latter requires building it, which is eager)

src/main/kotlin/org/pkl/lsp/VirtualFile.kt
```diff
@@ -197,12 +232,8 @@ class HttpsFile(override val uri: URI, override val project: Project) : BaseFile
...
   }
 
-  override fun toModule(): PklModule? {
-    return project.fileCacheManager.findHttpModule(uri)
-  }
-
-  override fun contents(): String {
-    return project.fileCacheManager.findHttpContent(uri) ?: ""
+  override fun doReadContents(): String {
+    return uri.toURL().readText()
```
stackoverflow, line 236: We had some checks before in case a https URL was not valid Pkl. I guess we'll just get generic parsing errors now.

stackoverflow (approved):
Much better now!

src/main/kotlin/org/pkl/lsp/VirtualFile.kt
bioball: This will throw `IOException` if the load fails, which we bubble up to `doBuildModule`, which then returns `null`.

It does mean that we will re-fetch HTTP every time, whereas `FileCacheManager` will fail on the first attempt. However, it's possible that a request might fail, but succeed later (no internet connection, etc), so I think retrying is a little better in this case.

Maybe we can throttle how often we retry, to prevent accidentally DOS'ing some poor internet server.

stackoverflow: Throttle is the best idea but that is an optimization we can do later.

bioball: Went ahead and added a debouncer (similar to throttle, but new calls push back the timer)

PR #17, opened 2024-09-16
switch parser to tree-sitter

stackoverflow:


src/main/kotlin/org/pkl/lsp/ast/PklNode.kt
```diff
@@ -741,112 +708,175 @@ abstract class AbstractPklNode(
...
+    text == "unknown" -> PklUnknownTypeImpl(project, parent!!, this)
+    text == "nothing" -> PklNothingTypeImpl(project, parent!!, this)
+    text == "module" -> PklModuleTypeImpl(project, parent!!, this)
+    childCount <= 0 -> null
+    children[0].type == "stringConstant" -> PklStringLiteralTypeImpl(project, parent!!, this)
+    children[0].type == "qualifiedIdentifier" -> PklDeclaredTypeImpl(project, parent!!, this)
+    children[0].type == "(" && childCount == 3 -> PklParenthesizedTypeImpl(project, parent!!, this)
+    children[0].type == "(" -> PklFunctionTypeImpl(project, parent!!, this)
+    // TODO: for pkl-tree-sitter `*Foo` alone is a valid type
+    children[0].type == "*" -> PklDefaultUnionTypeImpl(project, parent!!, this)
```
bioball, line 736: I think this is fine; we can add a diagnostic to flag this as an error through the language server. It might be even preferable to do it this way because we can provide a better error message and a quickfix.

.java-version
```diff
@@ -0,0 +1 @@
+22
```
bioball, line 1: My `jenv` doesn't have `22` as a version; it only has `22.0` and `22.0.2`. Maybe make this `22.0`?

stackoverflow: I have both in my `jenv`. Anyway, changed it to `22.0`.

src/main/kotlin/org/pkl/lsp/ast/Extensions.kt
```diff
@@ -172,27 +170,25 @@ private fun PklNode.getEscapedText(): String? = buildString {
...
-          } catch (ignored: NumberFormatException) {} catch (ignored: IllegalArgumentException) {}
+        if (text.contains("u{")) {
+          val index = text.indexOf('{') + 1
+          if (index != -1) {
+            val hexString = text.substring(index, text.length - 1)
+            try {
+              append(Character.toChars(Integer.parseInt(hexString, 16)))
+            } catch (ignored: NumberFormatException) {} catch (ignored: IllegalArgumentException) {}
+          }
+        } else {
```
holzensp, line 183: I'm not entirely sure how this worked before, but... `indexOf` returns `-1` when the pattern isn't found. I don't think it ever returns `-2`, does it? I guess it works because `indexOf` _never_ returns `-1`, considering this happens after we know `text.contains("u{")`. I'm also a _little_ thrown by `text.length - 1` for the end of the `hexString`; if we know the string has no trailing matter, why wasn't this a `text.startsWith("u{")` and `val index = 2` (which I'd expect if we believe `text` is "tight")?

.circleci/config.pkl
```diff
@@ -8,13 +8,18 @@ jobs {
...
     steps {
       "checkout"
       new RunStep {
-        command = "./gradlew --info --stacktrace -DtestReportsDir=${HOME}/test-results check"
+        command = "git submodule update --init --recursive"
+      }
+      new RunStep {
+        command = """
+          export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:. && ./gradlew --info --stacktrace -DtestReportsDir=${HOME}/test-results generateTreeSitterLib check
+          """
```
holzensp, line 22:
```suggestion
        command = """
          LD_LIBRARY_PATH="${LD_LIBRARY_PATH}:." ./gradlew --info --stacktrace -DtestReportsDir="${HOME}/test-results" generateTreeSitterLib check
          """
```

src/main/kotlin/org/pkl/lsp/analyzers/SyntaxAnalyzer.kt
```diff
@@ -17,21 +17,18 @@ package org.pkl.lsp.analyzers
...
-    if (del != null) {
-      val err =
-        if (del == ",") {
-          ErrorMessages.create("missingCommaSeparator")
-        } else {
-          ErrorMessages.create("missingDelimiter", del)
-        }
-      diagnosticsHolder += error(node, err)
+    if (node is PklError) {
+      diagnosticsHolder += error(node, ErrorMessages.create("invalidSyntax"))
```
holzensp, line 27: Not the _best_ error message; is there no way to reconstruct some extra details about expectations?

scripts/generate-tree-sitter-libs.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+
+# generate tree-sitter lib
+cd tree-sitter || exit
+make
+rm -rf "../src/main/resources/libtree-sitter.$1"
+cp "libtree-sitter.$1" ../src/main/resources
+cd ..
```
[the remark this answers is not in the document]
stackoverflow: This can definitely be improved! I just hacked this together to get the build green. I'm pretty sure we can move this whole script to gradle, it's just my gradle-fu was lacking.

src/main/kotlin/org/pkl/lsp/PklVisitor.kt
```diff
@@ -171,7 +171,7 @@ open class PklVisitor<R> {
     return visitElement(o)
   }
 
-  open fun visitModuleDeclaration(o: PklModuleDeclaration): R? {
+  open fun visitModuleDeclaration(o: PklModuleHeader): R? {
```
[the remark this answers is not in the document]
stackoverflow: Yes, good catch!

src/main/kotlin/org/pkl/lsp/ast/Extensions.kt
stackoverflow: You are right about the `if (index != -1)`. It won't ever be false because we already made sure with `contains` that this is a unicode escape.
About why `index` is not just `= 2`: we don't know how many `#` chars the string has so the escape may be `\###u{22}`.
Why the `text.length - 1`: because the escape ends with `}` which we have to remove to just get the number.
We could use a simple `\d+` regex to get just the number from the string, but using index is faster.

src/main/kotlin/org/pkl/lsp/analyzers/SyntaxAnalyzer.kt
stackoverflow: The whole syntax analyzer is in a pretty poor state and should be improved. Unfortunately the only thing the error node gives is its text, but that's already shown in the client.

PR #19, opened 2024-09-17
Render member links

bioball:
Support rendering member links in doc comments!

Member links: https://pkl-lang.org/main/current/language-reference/index.html#member-links

### DIFF
--- a/src/main/kotlin/org/pkl/lsp/documentation/DocCommentMemberLinkProcessor.kt
+++ b/src/main/kotlin/org/pkl/lsp/documentation/DocCommentMemberLinkProcessor.kt
@@ -1,0 +1,147 @@
+/**
+ * Copyright © 2024 Apple Inc. and the Pkl project authors. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.pkl.lsp.documentation
+
+import org.pkl.lsp.ast.*
+
+/**
+ * Resolve
+ * [member links](https://pkl-lang.org/main/current/language-reference/index.html#member-links), and
+ * turn them into editor links.
+ */
+object DocCommentMemberLinkProcessor {
+  private val linkRegex =
+    Regex(
+      """
+    (?x)            # turn on extended mode (newlines are ignored)
+    \[(.+?)]        # '[', followed by non-newline characters (1st capture), followed by ']'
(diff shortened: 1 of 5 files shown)


src/main/kotlin/org/pkl/lsp/documentation/DocCommentMemberLinkProcessor.kt
```diff
@@ -0,0 +1,129 @@
...
+      }
+    }
+  }
+
+  private fun resolveDocLink(text: String, ownerNode: PklNode): PklNode? {
+    var currentNode: PklNode? = ownerNode
+    val parts = text.split('.')
+    if (parts.isEmpty()) return null
+    val first = parts[0]
+    currentNode =
```
stackoverflow, line 74: Why not move this to the `for` and start from 0?

src/main/kotlin/org/pkl/lsp/documentation/DocCommentMemberLinkProcessor.kt
```diff
@@ -0,0 +1,129 @@
...
+      is PklClass ->
+        element.cache(context).methods[methodName] ?: element.parent?.resolveMethod(methodName)
+      is PklModule -> element.cache(context).methods[methodName]
+      else -> element.parent?.resolveMethod(methodName)
+    }
+  }
+
+  private fun PklNode.resolveVariable(variableName: String): PklNode? {
+    val context = containingFile.pklProject
+    return when (val element = this) {
```
stackoverflow, line 102: nit: you can just `when (this)`

src/main/kotlin/org/pkl/lsp/documentation/DocCommentMemberLinkProcessor.kt
```diff
@@ -0,0 +1,129 @@
...
+        when {
+          variableName == "module" -> this
+          else -> {
+            element.imports
+              .find { it.memberName == variableName }
+              ?.let { import ->
+                if (import.isGlob) import else import.resolveModules(context).first()
+              }
+              ?: element.cache(context).typeDefsAndProperties[variableName]
+              ?: project.pklBaseModule.module.cache(null).typeDefsAndProperties[variableName]
```
stackoverflow, line 117: Why not pass `context` here to the cache?

stackoverflow (approved):
LGTM

src/main/kotlin/org/pkl/lsp/documentation/DocCommentMemberLinkProcessor.kt
bioball: Good catch!

src/main/kotlin/org/pkl/lsp/documentation/DocCommentMemberLinkProcessor.kt
bioball: Ah, this was a bug. The 2nd - Nth lookups should resolve properties and tyedefs, but not other variables (e.g. no parameters). Will fix.

src/main/kotlin/org/pkl/lsp/documentation/DocCommentMemberLinkProcessor.kt
bioball: Providing context here will give us different cached members per context, which doesn't make sense for the base module because they are always the same.

PR #23, opened 2024-09-21
Improve handling of native libraries

bioball:
This implements changes to put native libs in `~/.pkl/editor-support/`. When loading the parser, it first looks for libraries there, and if they don't exist, they get copied from bundled resources.

When building libraries, the os name and architecture get included in the resource path.

This enables starting the jar plainly without needing any extra argument (e.g. no need for CLI flag `-Djava.library.path`, and no need for the `LD_LIBRARY_PATH` env var), and eliminates the need to copy native libraries to the current working dir.

Note: java-tree-sitter doesn't provide any way to provide the library path. To get around this, I'm doing a pretty ugly hack to rewrite the `TreeSitter.java` source file to make it load the tree-sitter library from our custom dir.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -15,12 +15,14 @@
  */
 import com.github.gradle.node.npm.task.NpmInstallTask
 import com.github.gradle.node.task.NodeTask
+import org.apache.tools.ant.filters.ReplaceTokens
 import org.gradle.internal.extensions.stdlib.capitalized
 import org.gradle.internal.os.OperatingSystem
 import org.jetbrains.kotlin.gradle.internal.ensureParentDirsCreated
 
 plugins {
   application
+  idea
   alias(libs.plugins.kotlin)
   alias(libs.plugins.kotlinSerialization)
   alias(libs.plugins.nodeGradle)
@@ -36,6 +38,15 @@ java {
 }
 
 val pklCli: Configuration by configurations.creating
+
+val jtreeSitterSources: Configuration by configurations.creating
+
+val buildInfo = extensions.create<BuildInfo>("buildInfo", project)
+
+val jsitterMonkeyPatchSourceDir = layout.buildDirectory.dir("generated/libs/jtreesitter")
+val nativeLibDir = layout.buildDirectory.dir("generated/libs/native/")
+val treeSitterPklRepoDir = layout.buildDirectory.dir("repos/tree-sitter-pkl")
+val treeSitterRepoDir = layout.buildDirectory.dir("repos/tree-sitter")
 
 val osName
   get(): String {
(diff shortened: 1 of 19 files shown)


src/main/kotlin/org/pkl/lsp/Project.kt
```diff
@@ -54,8 +53,7 @@ class Project(private val server: PklLSPServer) {
 
   val clientCapabilities: PklClientCapabilities by lazy { server.clientCapabilities }
 
-  // The thread where all tree-sitter allocations happen
-  val astExecutor: ExecutorService by lazy { Executors.newSingleThreadExecutor() }
+  val pklParser: PklParser by lazy { PklParser(this) }
```
bioball, line 56: Unrelated: make `PklParser` a component, and remove `astExecutor` because it's only used by the parser anyway.

build.gradle.kts
```diff
@@ -220,9 +262,30 @@ val makeTreeSitterPklLib by
...
+  // we can remove this when tree-sitter-pkl upgrades the tree-sitter-cli dependency to 0.23 or
+  // newer.
+  exclude("**/*.dSYM/**")
+  filesMatching("org/pkl/lsp/Release.properties") {
+    filter<ReplaceTokens>(
+      "tokens" to
+        mapOf(
+          "version" to buildInfo.pklLspVersion,
+          "treeSitterVersion" to libs.versions.treeSitterRepo.get(),
+          "treeSitterPklVersion" to libs.versions.treeSitterPklRepo.get(),
```
bioball, line 275: Provide versions of these libraries to runtime, because this affects where these libraries get loaded from. For example, on macOS, the tree-sitter lib lives at `~/.pkl/editor-support/native-libs/tree-sitter/v0.23.0/libtree-sitter.dylib`.

This ensures that the version of the native lib matches what the LSP expects.

buildSrc/src/main/kotlin/ExecutableJar.kt
```diff
@@ -0,0 +1,58 @@
...
+        .trimIndent() + "\n\n\n"
+    outFile.outputStream().use { outStream ->
+      startScript.byteInputStream().use { it.copyTo(outStream) }
+      inFile.inputStream().use { it.copyTo(outStream) }
+    }
+
+    // chmod a+x
+    outFile.setExecutable(true, false)
+  }
+}
```
bioball, line 58: Tangent: I introduced `BuildInfo` too, and moved both `BuildInfo` and `ExecutableJar` into `buildSrc` to mimick how we set things up in the pkl repo.

stackoverflow (approved):
LGTM

stackoverflow:
We should create an issue in java-tree-sitter so they support configurable paths for the tree-sitter library.
That hack is very ugly and brittle: we need to make sure it's still working every time we update java-tree-sitter.

src/main/kotlin/org/pkl/lsp/treesitter/NativeLibrary.kt
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  private val storedLibraryPath: Path by lazy {
+    nativeLibsDir.resolve("$name/$version/$systemLibraryName")
+  }
+
+  val libraryPath: Path by lazy {
+    when {
+      // optimization: if the resource file is a normal file, we can use it directly.
+      resourcePath.fileSystem == FileSystems.getDefault() -> resourcePath
```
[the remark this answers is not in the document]
bioball: Yup

PR #24, opened 2024-09-23
Use Zig to cross-compile native libs

bioball:
This uses Zig to compile the native libraries for macOS/linux/windows for the aarch64/amd64 architectures.

This also removes Node.js stuff, which is no longer necessary because we don't use the tree-sitter CLI anymore to build these libraries.

### DIFF
--- a/buildSrc/src/main/kotlin/BuildInfo.kt
+++ b/buildSrc/src/main/kotlin/BuildInfo.kt
@@ -20,10 +20,11 @@ import org.gradle.api.artifacts.VersionCatalog
 import org.gradle.api.artifacts.VersionCatalogsExtension
 import org.gradle.kotlin.dsl.getByType
 import org.gradle.internal.os.OperatingSystem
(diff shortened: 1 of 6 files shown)


buildSrc/src/main/kotlin/BuildInfo.kt
```diff
@@ -73,6 +83,16 @@ open class BuildInfo(project: Project) {
     project.extensions.getByType<VersionCatalogsExtension>().named("libs")
   }
 
+  val zig: Zig = Zig()
+
+  inner class Zig {
+    val version: String get() = libs.findVersion("zig").get().toString()
+
+    val installDir: Path get() = project.projectDir.toPath().resolve(".gradle/zig/zig-$version")
```
bioball, line 91: I followed the node-gradle plugin's lead and put the Zig distribution in `.gradle`. Not sure if this is the correct thing though.

stackoverflow (approved):
Looks good

PR #25, opened 2024-09-24
add unqualified completions

stackoverflow:


src/main/kotlin/org/pkl/lsp/completion/UnqualifiedAccessCompletionProvider.kt
```diff
@@ -0,0 +1,60 @@
...
+      null,
+      isProperty = false,
+      project.pklBaseModule,
+      mapOf(),
+      visitor,
+      context,
+    )
+    collector.addAll(visitor.result)
+  }
+}
```
bioball, line 60: There's a fair bit of logic that exists in our pkl-intellij version that's missing here: https://github.com/apple/pkl-intellij/blob/main/src/main/kotlin/org/pkl/intellij/completion/UnqualifiedAccessCompletionProvider.kt

Is that planned as a follow-up? Or do you think there should be a separate `DefinitionCompletionProvider`?

src/main/kotlin/org/pkl/lsp/resolvers/ResolveVisitors.kt
```diff
@@ -426,14 +426,19 @@ object ResolveVisitors {
...
+    prefix: String? = null,
+  ): ResolveVisitor<Set<CompletionItem>> {
+    return object : ResolveVisitor<Set<CompletionItem>> {
       override fun visit(
         name: String,
         element: PklNode,
         bindings: TypeParameterBindings,
         context: PklProject?,
       ): Boolean {
+        if (prefix != null && !name.startsWith(prefix)) return true
```
bioball, line 440: Probably better to just send over all completions and let the IDE present each of these options.

The editor should use this to let users fuzzy-find completion members (and even matches on names with mismatched cases)

Screenshot 2024-09-24 at 7 21 54 AM


```suggestion
```

bioball (approved):
Approving to unblock, but, see comments!

src/main/kotlin/org/pkl/lsp/resolvers/ResolveVisitors.kt
stackoverflow: You are right, I somehow missed that the client would do the filtering.

src/main/kotlin/org/pkl/lsp/completion/UnqualifiedAccessCompletionProvider.kt
```diff
@@ -0,0 +1,265 @@
...
+    node: PklNode,
+    params: CompletionParams,
+    collector: MutableList<CompletionItem>,
+  ) {
+    val line = params.position.line + 1
+    val column = params.position.character + 1
+    val context = node.containingFile.pklProject
+    val base = project.pklBaseModule
+    val actualNode =
+      node.enclosingModule?.findBySpan(line, column) as? PklUnqualifiedAccessExpr ?: return
```
bioball, line 46: I think we'll need something a little more sophisticated, because it can't handle cases like:

```
foo = <cursor>
```

```
foo {
  <cursor>
}
```

```
foo = let (bar = <cursor>) bar + 5
```

Maybe one approach here could be to: inject an expression, then see if the cursor position becomes unqualified access?

PR #27, opened 2024-09-24
Use inputs.property instead of manually tracking

bioball:
I realized that Gradle provides a way to use a simple property to help track whether something is up-to-date, so it was totally unnecessary to create one-off files ourselves.


build.gradle.kts
```diff
@@ -158,18 +158,8 @@ fun configureRepo(
     }
 
   return tasks.register("setup$taskSuffix") {
-    dependsOn(cloneTask)
     dependsOn(updateTask)
```
bioball, line 161: Also: this fixes a race condition when executing with parallel mode (e.g. `./gradlew build --parallel`)

stackoverflow (approved):

PR #28, opened 2024-09-30
Don't throw exception on errors and comments inside binary expressions

gordonbondon:


src/test/kotlin/org/pkl/lsp/ParserTest.kt
```diff
@@ -196,6 +196,41 @@ class ParserTest {
...
+
+    val mod = parse(code)
+    val prop = mod.children[0]
+    assertThat(prop).isInstanceOf(PklClassProperty::class.java)
+    prop as PklClassProperty
+    assertThat(prop.type).isNull()
+  }
+
+  @Test
+  fun `Nested error inside binary expression are parsed correctly`() {
```
holzensp, line 218: Nit
```suggestion
  fun `Nested errors inside binary expression are parsed correctly`() {
```

src/test/kotlin/org/pkl/lsp/ParserTest.kt
```diff
@@ -196,6 +196,41 @@ class ParserTest {
...
+        || bar.]
+        || three
+    """
+        .trimIndent()
+
+    val mod = parse(code)
+    val prop = mod.children[0]
+    assertThat(prop).isInstanceOf(PklClassProperty::class.java)
+    prop as PklClassProperty
+    assertThat(prop.type).isNull()
```
holzensp, line 231: nit
```suggestion
    assertThat(prop)
        .isInstanceOf(PklClassProperty::class.java)
        .hasFieldOrPropertyWithValue("type", null)
```

holzensp (approved):
LGTM; leaving to @stackoverflow to sign-off/merge. Couple aesthetics nits.

src/main/kotlin/org/pkl/lsp/ast/PklNode.kt
```diff
@@ -824,6 +824,8 @@ fun TreeSitterNode.toNode(project: Project, parent: PklNode?): PklNode? {
       when (children[1].type) {
         "**" -> PklExponentiationExprImpl(project, parent!!, this)
         "??" -> PklNullCoalesceExprImpl(project, parent!!, this)
+        "lineComment",
+        "ERROR" -> PklErrorImpl(project, parent!!, this)
```
stackoverflow, line 828: That's definitely a bug! But there are some problems with your solution:
- `lineComment` shouldn't become an error, otherwise the LSP will report a syntax error on the comment
- `blockComment` will still thrown an exception in cases like `foo /* asd */ || bar`

The problem here is the `when (children[1].type)` which assumes the first child is the operator, but that's not always true. We need better heuristics.

src/main/kotlin/org/pkl/lsp/ast/PklNode.kt
```diff
@@ -843,6 +845,8 @@ fun TreeSitterNode.toNode(project: Project, parent: PklNode?): PklNode? {
         "&&" -> PklLogicalAndExprImpl(project, parent!!, this)
         "||" -> PklLogicalOrExprImpl(project, parent!!, this)
         "|>" -> PklPipeExprImpl(project, parent!!, this)
+        "lineComment",
+        "ERROR" -> PklErrorImpl(project, parent!!, this)
```
stackoverflow, line 849: Same as the comment above.

src/main/kotlin/org/pkl/lsp/ast/PklNode.kt
gordonbondon: Addressed handling of `blockComment`

stackoverflow (approved):

PR #29, opened 2024-09-30
Update tree-sitter-pkl to latest version

bioball:
This updates tree-sitter-pkl, and updates logic accordingly.

### DIFF
--- a/src/main/kotlin/org/pkl/lsp/ast/Terminal.kt
+++ b/src/main/kotlin/org/pkl/lsp/ast/Terminal.kt
@@ -151,9 +151,9 @@ fun TreeSitterNode.toTerminal(parent: PklNode): Terminal? {
       "identifier" -> TokenType.Identifier
       "docComment" -> TokenType.DocComment
       "escapeSequence" -> TokenType.CharacterEscape
+      "\n" -> TokenType.Whitespace
       // TODO: see if we need these
       //      PklParser.NewlineSemicolon -> TokenType.NewlineSemicolon
-      //      PklParser.Whitespace -> TokenType.Whitespace
       //      PklParser.BlockComment -> TokenType.BlockComment
       //      PklParser.LineComment -> TokenType.LineComment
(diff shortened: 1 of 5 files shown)


src/main/kotlin/org/pkl/lsp/ast/Terminal.kt
```diff
@@ -151,9 +151,9 @@ fun TreeSitterNode.toTerminal(parent: PklNode): Terminal? {
       "identifier" -> TokenType.Identifier
       "docComment" -> TokenType.DocComment
       "escapeSequence" -> TokenType.CharacterEscape
+      "\n" -> TokenType.Whitespace
```
bioball, line 154: This now gets emitted as a token because `\n` is explicitly provided as an alternative for `propertyCallExpr`.

src/test/kotlin/org/pkl/lsp/ParserTest.kt
```diff
@@ -196,6 +196,28 @@ class ParserTest {
...
+      }
+    """
+        .trimIndent()
+    val mod = parse(code)
+    val prop = mod.children[0]
+    assertThat(prop).isInstanceOf(PklClassProperty::class.java)
+    prop as PklClassProperty
+    val whenGenerator = prop.objectBody!!.members.first() as PklWhenGenerator
+    assertThat(whenGenerator.thenBody).isNotNull
+    assertThat(whenGenerator.elseBody).isNotNull
```
stackoverflow, line 218:
```suggestion
    assertThat(whenGenerator.thenBody).isInstanceOf(PklObjectBody::class.java)
    assertThat(whenGenerator.elseBody).isInstanceOf(PklObjectBody::class.java)
```
nit: a bit stronger check.

stackoverflow (approved):
Looks good

src/test/kotlin/org/pkl/lsp/ParserTest.kt
bioball: This is already ensured by the compiler!

PR #31, opened 2024-10-04
Set up CI stuff, get rid of ExecutableJar

bioball:
This changes our logic around publishing:

* CircleCI: build shadowJar, then verify it on macOS/windows (linux gets verified during the build), before publishing to maven
* Set up maven publication with signing
* Set up release job and main (snapshot) job
* Remove ExecutableJar; since the LSP is not a user-facing tool, it does not need to be in an executable jar format.
* Fix building of native libraries; it was using the wrong flag


stackoverflow (approved):
LGTM

PR #32, opened 2024-10-04
fix binary expressions AST generation

stackoverflow:
While generating the AST for binary expressions we were making wrong assumptions about where the children should be located. This fixes the generation so expressions with comments, errors, or any other kind of node in between will still parse correctly.

### DIFF
--- a/src/main/kotlin/org/pkl/lsp/ast/Expr.kt
+++ b/src/main/kotlin/org/pkl/lsp/ast/Expr.kt
@@ -15,6 +15,7 @@
  */
 package org.pkl.lsp.ast
 
+import java.util.EnumSet
 import org.pkl.lsp.*
 import org.pkl.lsp.LSPUtil.firstInstanceOf
 import org.pkl.lsp.packages.dto.PklProject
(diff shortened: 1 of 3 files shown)


holzensp (approved):
LGTM

src/main/kotlin/org/pkl/lsp/ast/Expr.kt
```diff
@@ -400,15 +401,42 @@ class PklLogicalNotExprImpl(
   }
 }
 
-class PklAdditiveExprImpl(
+abstract class PklBinExprImpl(
   override val project: Project,
   override val parent: PklNode,
   override val ctx: TreeSitterNode,
```
bioball, line 407: Let's add `override val operator: PklNode` as a constructor arg here, since we've already found the node when creating this tree.

src/main/kotlin/org/pkl/lsp/ast/PklNode.kt
```diff
@@ -824,17 +824,19 @@ fun TreeSitterNode.toNode(project: Project, parent: PklNode?): PklNode? {
...
-        "lineComment",
-        "blockComment" -> null
-        "ERROR" -> PklErrorImpl(project, parent!!, this)
-        else -> throw RuntimeException("Unknown binary operator `${children[1].type}`")
+        else -> throw RuntimeException("Unknown binary operator in expr `$text`")
       }
-    "binaryExpr" ->
-      when (children[1].type) {
+    }
+    "binaryExpr" -> {
```
bioball, line 835: Nit: you can combine `"binaryExprRightAssoc"` and `"binaryExpr"` into the same block, since they basically have the same logic.

bioball (approved):
Approving to unblock, but, see comment!

PR #33, opened 2024-10-04
Introduce ParserSnippetTests

bioball:
This introduces parser snippet tests, which parses input text, and writes out the AST of the parse tree.

This also moves some existing tests from `ParserTest` to `ParserSnippetTests`.


stackoverflow (approved):
Nice work!

PR #40, opened 2024-10-09
add definition-level completions

stackoverflow:


src/main/kotlin/org/pkl/lsp/completion/UnqualifiedAccessCompletionProvider.kt
```diff
@@ -239,48 +256,391 @@ class UnqualifiedAccessCompletionProvider(private val project: Project) : Comple
...
+              is PklModule -> {
+                collector.addAll(MODULE_LEVEL_KEYWORD_ELEMENTS)
+              }
+              is PklClass -> {
+                collector.addAll(CLASS_LEVEL_KEYWORD_ELEMENTS)
+              }
+              else -> {}
+            }
+          }
+        }
```
holzensp, line 340: Feels like a case where `if` might actually be quicker _and_ cleaner... Not blocking, just saying.

holzensp, line 340: Feels like a case where `if` might actually less branchy... Not blocking, just saying.

src/main/kotlin/org/pkl/lsp/completion/UnqualifiedAccessCompletionProvider.kt
```diff
@@ -239,48 +256,391 @@ class UnqualifiedAccessCompletionProvider(private val project: Project) : Comple
...
+    val amendedValueType = valueType.amended(base, context)
+    if (amendedValueType != Type.Nothing && amendedValueType != Type.Unknown) {
+      val amendingValueType = valueType.amending(base, context)
+      if ("default" !in alreadyDefinedProperties) {
+        collector +=
+          createPropertyAmendElement("default", amendingValueType, base.mappingDefaultProperty)
+      }
+      collector += createEntryAmendElement(keyType, amendingValueType, base)
+    }
+    if ("default" !in alreadyDefinedProperties) {
```
holzensp, line 417: This smells off... do you want to overwrite the `amendingValueType`'s `default` here?

src/main/kotlin/org/pkl/lsp/completion/UnqualifiedAccessCompletionProvider.kt
```diff
@@ -239,48 +256,391 @@ class UnqualifiedAccessCompletionProvider(private val project: Project) : Comple
...
+    val amendedElementType = elementType.amended(base, context)
+    if (amendedElementType != Type.Nothing && amendedElementType != Type.Unknown) {
+      val amendingElementType = elementType.amending(base, context)
+      if ("default" !in alreadyDefinedProperties) {
+        collector +=
+          createPropertyAmendElement("default", amendingElementType, base.listingDefaultProperty)
+      }
+      collector += createPropertyAmendElement("new", amendingElementType, null)
+    }
+    if ("default" !in alreadyDefinedProperties) {
```
holzensp, line 442: ^^ ditto

holzensp (approved):
Overall, LGTM; one point of doubt. Approving to unblock.

src/main/kotlin/org/pkl/lsp/completion/UnqualifiedAccessCompletionProvider.kt
stackoverflow: Simplified the conditions.

src/main/kotlin/org/pkl/lsp/completion/UnqualifiedAccessCompletionProvider.kt
stackoverflow: Not sure what you mean by "overwrite the `amendingValueType`'s `default`".

This code will add a `default = ...` autocomplete suggestion, for the case where you want to overwrite the `Mapping`/`Listing` default.

src/main/kotlin/org/pkl/lsp/completion/UnqualifiedAccessCompletionProvider.kt
```diff
@@ -239,48 +256,382 @@ class UnqualifiedAccessCompletionProvider(private val project: Project) : Comple
...
+  }
+
+  private fun createPropertyAssignElement(
+    propertyName: String,
+    propertyType: Type,
+    propertyCtx: PklProperty,
+  ): CompletionItem {
+    return CompletionItem("$propertyName = ").apply {
+      insertTextFormat = InsertTextFormat.Snippet
+      insertText = "${propertyCtx.modifiersStr}$propertyName = \${1:value}"
```
bioball, line 464: Here, and everywhere else: I think the placeholder text adds a little more noise than is helpful.

```suggestion
      insertText = "${propertyCtx.modifiersStr}$propertyName = $1"
```

bioball (changes requested):
This is looking pretty decent! I have some minor comments.

src/main/kotlin/org/pkl/lsp/completion/UnqualifiedAccessCompletionProvider.kt
```diff
@@ -65,9 +85,6 @@ class UnqualifiedAccessCompletionProvider(private val project: Project) : Comple
       }
```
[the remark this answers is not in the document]
stackoverflow: Ah yeah, because we added a character to the source. Good catch!

src/main/kotlin/org/pkl/lsp/completion/UnqualifiedAccessCompletionProvider.kt
```diff
@@ -239,48 +256,382 @@ class UnqualifiedAccessCompletionProvider(private val project: Project) : Comple
...
+
+    for ((propertyName, property) in properties) {
+      if (propertyName in alreadyDefinedProperties) continue
+      if (property.isFixedOrConst && !isClassOrModule) continue
+
+      val propertyType = property.type.toType(base, thisType.bindings, context)
+      val amendedPropertyType = propertyType.amended(base, context)
+      if (amendedPropertyType != Type.Nothing && amendedPropertyType != Type.Unknown) {
+        val amendingPropertyType = propertyType.amending(base, context)
+        collector += createPropertyAmendElement(propertyName, amendingPropertyType, property)
```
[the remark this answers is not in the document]
stackoverflow: Will the `amended` type always be the same as the `amending` type here?
We have the same logic in `addMappingCompletions` and `addListingCompletions`.

src/main/kotlin/org/pkl/lsp/completion/UnqualifiedAccessCompletionProvider.kt
```diff
@@ -239,48 +256,382 @@ class UnqualifiedAccessCompletionProvider(private val project: Project) : Comple
...
+        """
+        ${propertyCtx.modifiersStr}$propertyName {
+          ${'$'}{1:body}
+        }
+        """
+          .trimIndent()
+      detail = propertyType.render()
+      kind = CompletionItemKind.Property
+    }
+  }
```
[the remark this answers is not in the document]
stackoverflow: Added `completionItem` to `PklProperty`.

src/main/kotlin/org/pkl/lsp/ast/PklNode.kt
```diff
@@ -258,6 +260,16 @@ sealed interface PklProperty :
   val expr: PklExpr?
   val isDefinition: Boolean
   val typeAnnotation: PklTypeAnnotation?
+
+  fun toCompletionItem(): CompletionItem {
+    val item = CompletionItem(name)
+    item.kind = CompletionItemKind.Field
+    item.detail = type?.render() ?: "unknown"
+    if (this is PklClassProperty) {
+      item.documentation = getDoc(this, containingFile.pklProject)
```
bioball, line 269: We can actually get rid of `effectiveDocComment`. I think I created it but forgot to get rid of it.

This can just be:

```suggestion
      item.documentation = parsedComment
```

Also, style-wise, I like using `apply` as a builder, e.g.

```
return CompletionItem(name).apply {
  kind = CompletionItemKind.Field
  // etc
}
```

bioball (approved):
LGTM!

PR #41, opened 2024-10-09
Add documentation

bioball:
This adds an antora documentation component!

### DIFF
--- a/docs/antora.yml
+++ b/docs/antora.yml
@@ -1,0 +1,5 @@
+name: lsp
+title: pkl-lsp
+version: 0.1.1
+nav:
+  - nav.adoc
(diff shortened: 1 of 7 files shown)


docs/antora.yml
```diff
@@ -0,0 +1,5 @@
+name: lsp
+title: pkl-lsp
+version: 0.1.1
```
stackoverflow, line 3: Current version is 0.1.0

stackoverflow (approved):

docs/antora.yml
bioball: Yeah, but the version on main is 0.1.1 (see https://github.com/apple/pkl-lsp/blob/689fb86fa24bdf2aa62d170b491704987045ca0c/gradle.properties#L2)

PR #44, opened 2024-10-10
Allow access to tree-sitter trees from multiple threads

stackoverflow:
Tree-sitter trees are safe to be read from other threads ([context](https://github.com/tree-sitter/tree-sitter/issues/2612#issuecomment-1711960164)).
This PR patches the code so we don't need to funnel all AST reads through a single thread.


build.gradle.kts
```diff
@@ -131,6 +136,14 @@ tasks.test {
   }
 }
 
+tasks.distZip { dependsOn(tasks.shadowJar) }
+
+tasks.distTar { dependsOn(tasks.shadowJar) }
+
+tasks.named("startScripts") { dependsOn(tasks.shadowJar) }
+
+tasks.named("startShadowScripts") { dependsOn(tasks.jar) }
```
stackoverflow, line 145: Had to add these lines or `./gradlew build` wouldn't work for me.

bioball: Shouldn't this one be:

```suggestion
tasks.named("startShadowScripts") { dependsOn(tasks.shadowJar) }
```

build.gradle.kts
```diff
@@ -102,6 +103,10 @@ val monkeyPatchTreeSitter by
             import org.pkl.lsp.treesitter.NativeLibraries;
           """
               .trimIndent()
+          line.contains("Arena.ofConfined") -> line.replace("ofConfined", "ofShared")
+          line.contains("jspecify") -> ""
+          line.contains("@NullMarked") -> ""
+          line.contains("@Nullable") -> line.replace("@Nullable ", "")
```
bioball, line 109: What's these ones about?

bioball (approved):
Approving to unblock, some questions

build.gradle.kts
stackoverflow: I added the exact dependencies that gradle complained should be there. You can try to reproduce it locally with `./gradlew build`. @holzensp had the same error also.

build.gradle.kts
bioball: Ah, it provides compile-time annotations. Gotcha.

build.gradle.kts
bioball: Some of these come from the application plugin, which we can probably just get rid of, since we're not using it. But that can be in another PR.

PR #51, opened 2024-11-06
Upgrade to java-tree-sitter 0.24

stackoverflow:
This version fixes two problems where we had to make workarounds:

- `Node.getText()` was not encode-aware.
- `Tree`s were not allowed to be shared between threads.

This PR upgrades the version and removes the workarounds.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -85,13 +85,12 @@ idea { module { generatedSourceDirs.add(jsitterMonkeyPatchSourceDir.get().asFile
  * `java.library.path`.
  *
  * This patches its source code so that we can control exactly where the tree-sitter library
- * resides. It also patches the `Tree` class so its objects can be accessed from multiple threads.
- * As we never modify trees, they are safe to be accessed.
+ * resides.
  */
 val monkeyPatchTreeSitter by
   tasks.registering(Copy::class) {
     from(zipTree(jtreeSitterSources.singleFile)) {
-      include("**/TreeSitter.java", "**/Tree.java")
+      include("**/TreeSitter.java")
       filter { line ->
         when {
           line.contains("static final SymbolLookup") ->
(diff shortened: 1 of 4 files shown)


build.gradle.kts
```diff
@@ -85,13 +85,12 @@ idea { module { generatedSourceDirs.add(jsitterMonkeyPatchSourceDir.get().asFile
...
  * This patches its source code so that we can control exactly where the tree-sitter library
- * resides. It also patches the `Tree` class so its objects can be accessed from multiple threads.
- * As we never modify trees, they are safe to be accessed.
+ * resides.
  */
 val monkeyPatchTreeSitter by
   tasks.registering(Copy::class) {
     from(zipTree(jtreeSitterSources.singleFile)) {
-      include("**/TreeSitter.java", "**/Tree.java")
+      include("**/TreeSitter.java")
```
holzensp, line 93: Quite minor, but the `filter` below can now be a simple `if` instead of a `when` (commenting here, because GH-PR-yippee)

src/main/kotlin/org/pkl/lsp/ast/PklNode.kt
```diff
@@ -717,7 +717,7 @@ abstract class AbstractPklNode(
 
   override val children: List<PklNode> by lazy { childrenByType.values.flatten() }
 
-  override val text: String by lazy { ctx.utfAwareText() }
+  override val text: String by lazy { ctx.text ?: "" }
```
holzensp, line 720: Shouldn't the `null` case still `throw`?

holzensp (approved):
No blockers, but do double check the now-no-longer-throwing behaviour, please.

src/main/kotlin/org/pkl/lsp/ast/PklNode.kt
stackoverflow: In practice this won't ever happen because we always create trees via a source String. But yes, in case this is ever `null` it's better to throw. I removed the default so it will null-pointer.

PR #57, opened 2024-12-04
Tweak Gradle build

odenix:
- fix IntelliJ errors in build.gradle.kts by declaring `buildInfo`
    - distinguish between jdkVersion and jdkTargetVersion and set them once in BuildInfo
      - it's often preferable to build with the latest JDK version even
        if targeting an older version
    - enforce jdkTargetVersion in compilation tasks
      - prevents accidental use of newer JDK APIs
    - run tests with `--enable-native-access=ALL-UNNAMED` to prevent warnings


bioball (approved):
LGTM, thank you!

PR #74, opened 2025-04-03
Miscellaneous fixes

bioball:
This fixes various issues in the LSP

* Fixes parsing of modifier nodes
* Fixes an issue where child nodes are constructed twice (`toNode()`, `toTerminal()` should never be called in the body of a class because they are already initialized when `children` is called)
* Improves child node selection by using field names
* Fixes an issue where stdlib modules can possibly be initialized multiple times
* Fixes possible race conditions to prevent a cached value's provider from being called multiple times
* Treat `PklMethod` as an identifier owner, so that code actions that target them can refer to them by name.
* Cache the result of `computeExprType`

### DIFF
--- a/src/main/kotlin/org/pkl/lsp/PklTextDocumentService.kt
+++ b/src/main/kotlin/org/pkl/lsp/PklTextDocumentService.kt
@@ -51,9 +51,6 @@ class PklTextDocumentService(project: Project) : Component(project), TextDocumen
 
   override fun didOpen(params: DidOpenTextDocumentParams) {
     val uri = URI(params.textDocument.uri)
-    project.virtualFileManager.get(uri)?.let { file ->
-      file.version = params.textDocument.version.toLong()
-    }
     project.messageBus.emit(textDocumentTopic, TextDocumentEvent.Opened(uri))
   }
 
--- a/src/main/kotlin/org/pkl/lsp/VirtualFileManager.kt
+++ b/src/main/kotlin/org/pkl/lsp/VirtualFileManager.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2024 Apple Inc. and the Pkl project authors. All rights reserved.
+ * Copyright © 2024-2025 Apple Inc. and the Pkl project authors. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,10 +17,10 @@ package org.pkl.lsp
 
 import java.net.URI
 import java.nio.file.*
-import java.util.*
+import java.util.concurrent.ConcurrentHashMap
 
 class VirtualFileManager(project: Project) : Component(project) {
-  private val files: MutableMap<URI, BaseFile> = Collections.synchronizedMap(WeakHashMap())
+  private val files: MutableMap<URI, BaseFile> = ConcurrentHashMap()
 
   fun get(path: Path): VirtualFile? {
     return if (!Files.exists(path)) null else get(path.toUri(), path)
--- a/src/main/kotlin/org/pkl/lsp/ast/Expr.kt
+++ b/src/main/kotlin/org/pkl/lsp/ast/Expr.kt
@@ -267,18 +267,8 @@ class PklAmendExprImpl(
   override val parent: PklNode,
   override val ctx: Node,
 ) : AbstractPklNode(project, parent, ctx), PklAmendExpr {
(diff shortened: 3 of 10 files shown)


src/main/kotlin/org/pkl/lsp/PklTextDocumentService.kt
```diff
@@ -51,9 +51,6 @@ class PklTextDocumentService(project: Project) : Component(project), TextDocumen
 
   override fun didOpen(params: DidOpenTextDocumentParams) {
     val uri = URI(params.textDocument.uri)
-    project.virtualFileManager.get(uri)?.let { file ->
-      file.version = params.textDocument.version.toLong()
-    }
```
bioball, line 56: This logic was causing the language server to initialize multiple versions of the standard library, because we initialize it upon startup with a version of `-1`, but we then set it to `1` when the client opens up a stdlib module in the editor.

src/main/kotlin/org/pkl/lsp/VirtualFileManager.kt
```diff
@@ -17,10 +17,10 @@ package org.pkl.lsp
 
 import java.net.URI
 import java.nio.file.*
-import java.util.*
+import java.util.concurrent.ConcurrentHashMap
 
 class VirtualFileManager(project: Project) : Component(project) {
-  private val files: MutableMap<URI, BaseFile> = Collections.synchronizedMap(WeakHashMap())
+  private val files: MutableMap<URI, BaseFile> = ConcurrentHashMap()
```
bioball, line 23: This was yet another place that was causing us to initialize the stdlib multiple times. The weakmap was kicking out modules (including stdlib modules).

Since modules can point to other modules directly (e.g. via computing a node's type), I'm not really sure if using a weak map is actually the correct thing to do.

This does mean that the server's memory will grow. However, in practice, I'm not sure if this matters all that much. For example, many editor tools will scan and index an entire project and create symbols, which means that they have every project file stored in an internal state like we will here.

src/main/kotlin/org/pkl/lsp/ast/Expr.kt
```diff
@@ -380,8 +370,10 @@ class PklSubscriptExprImpl(
   override val parent: PklNode,
   override val ctx: Node,
 ) : AbstractPklNode(project, parent, ctx), PklSubscriptExpr {
-  override val leftExpr: PklExpr by lazy { ctx.children[0].toNode(project, this) as PklExpr }
-  override val rightExpr: PklExpr by lazy { ctx.children[2].toNode(project, this) as PklExpr }
```
bioball, line 384: This was the source of some bugs: we already initialize a node's children, but then initialize them yet again with these `toNode()` calls. This means that we have duplicate representations, which is both a memory/cpu waste, and also was messing with our resolvers, which use `===` to compare nodes.

src/main/kotlin/org/pkl/lsp/ast/ModuleOrClass.kt
```diff
@@ -143,7 +143,7 @@ class PklModuleClauseImpl(
     getChild(PklQualifiedIdentifierImpl::class)
   }
 
-  override val modifiers: List<Terminal>? by lazy { terminals.takeWhile { it.isModifier } }
+  override val modifiers: List<Terminal>? by lazy { terminals.filter { it.isModifier } }
```
bioball, line 146: This is incorrect because the first node might be a line comment, for example.

src/main/kotlin/org/pkl/lsp/VirtualFileManager.kt
KushalP: If this is a concern you could turn this into a `ConcurrentMap<URI, WeakReference<BaseFile>>`, with an adjacent `ConcurrentLinkedQueue<URI>()` to track loading/unloading of file content as needed.

KushalP (approved):
Approving as the code looks fine, but it would be useful to have (regression) tests to capture these cases if at all possible:

- Fixes parsing of modifier nodes
- Fixes an issue where child nodes are constructed twice (toNode(), toTerminal() should never be called in the body of a class because they are already initialized when children is called)
- Fixes an issue where stdlib modules can possibly be initialized multiple times
- Fixes possible race conditions to prevent a cached value's provider from being called multiple times

KushalP (approved):

bioball:
Most of these are kind of tricky to test for. I'll add a test around the parsing of modifiers, though!

KushalP (approved):

PR #76, opened 2025-04-07
Add typechecking for class properties

bioball:
This fixes an issue where amending class properties are not typechecked.

### DIFF
--- a/src/main/kotlin/org/pkl/lsp/ast/Member.kt
+++ b/src/main/kotlin/org/pkl/lsp/ast/Member.kt
@@ -51,7 +51,34 @@ class PklClassPropertyImpl(
 
   override val type: PklType? by lazy { typeAnnotation?.type }
 
-  override val isDefinition: Boolean by lazy { expr != null }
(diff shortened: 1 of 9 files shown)


src/main/kotlin/org/pkl/lsp/ast/Member.kt
```diff
@@ -51,7 +51,34 @@ class PklClassPropertyImpl(
...
+            }
+            owner.supermodule(context)?.let { supermodule ->
+              return !supermodule.cache(context).properties.containsKey(name)
+            }
+            true
+          }
+          else -> false
+        }
+      }
+    }
```
bioball, line 81: This implementation is copied from https://github.com/apple/pkl-intellij/blob/255491bc6ced4d653bf31fa17e8935203db7e3f5/src/main/kotlin/org/pkl/intellij/psi/PklClassPropertyBase.kt#L52

HT154 (approved):
LGTM, one question re: tests

PR #80, opened 2025-04-09
Add analysis for type names

bioball:
This adds diagnostics like "Unresolved reference 'Foo'" when used as a type name.

As part of this, the parse tree is also fixed to better match the IntelliJ parse tree.

### DIFF
--- a/src/main/kotlin/org/pkl/lsp/analyzers/TypeNameAnalyzer.kt
+++ b/src/main/kotlin/org/pkl/lsp/analyzers/TypeNameAnalyzer.kt
@@ -1,0 +1,57 @@
+/*
+ * Copyright © 2025 Apple Inc. and the Pkl project authors. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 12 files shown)


src/main/kotlin/org/pkl/lsp/analyzers/TypeNameAnalyzer.kt
```diff
@@ -0,0 +1,57 @@
...
+              node.simpleTypeName,
+              ErrorMessages.create("unresolvedReference", identifier.text),
+            )
+          }
+          return false
+        }
+      }
+    }
+    return true
+  }
```
KushalP, line 56: As we're returning early here, and not accumulating `unresolvedReference` errors, I'm wondering if this should turn into a `when` expression:

```suggestion
  override fun doAnalyze(node: PklNode, holder: DiagnosticsHolder): Boolean =
    when {
      node is PklTypeName -> {
        val context = node.containingFile.pklProject
        when {
          node.moduleName != null && node.moduleName!!.resolve(context) == null -> {
            node.moduleName?.identifier?.let {
              holder.addError(
                node.moduleName!!,
                ErrorMessages.create("unresolvedReference", it.text),
              )
            }
            false
          }
          node.simpleTypeName.resolve(context) == null -> {
            node.simpleTypeName.identifier?.let {
              holder.addError(
                node.simpleTypeName,
                ErrorMessages.create("unresolvedReference", it.text),
              )
            }
            false
          }
          else -> true
        }
      }
      else -> true
    }
```

KushalP (approved):
One minor suggestion, which you can take/leave.

PR #82, opened 2025-04-10
Add analysis for unresolved class/object properties; missing modifier 'local'

bioball:
This adds analysis for unresolved class/object properties.

Additionally, it adds analysis for more members that are missing the 'local' modifier.

As part of this change, the `ModuleMemberAnalyzer` is changed to simply `MemberAnalyzer`, because much of the logic can be shared between the two.


src/main/kotlin/org/pkl/lsp/analyzers/MemberAnalyzer.kt
```diff
@@ -0,0 +1,103 @@
...
+/** Analyzes object/class members, except for modifiers. */
+class MemberAnalyzer(project: Project) : Analyzer(project) {
+  override fun doAnalyze(node: PklNode, holder: DiagnosticsHolder): Boolean {
+
+    val module = node.enclosingModule ?: return false
+    val project = module.project
+    val base = project.pklBaseModule
+    val context = module.containingFile.pklProject
+
+    val memberType: Type by lazy {
```
stackoverflow, line 43: Why make it lazy when you are forcing it a few lines below?

stackoverflow (approved):

src/main/kotlin/org/pkl/lsp/analyzers/MemberAnalyzer.kt
bioball: We only execute this if the node is `PklObjectProperty` or `PklClassProperty`

PR #92, opened 2025-05-19
Fix deadlock

bioball:
This fixes a deadlock that arises when computing diagnostics.

This also generally reduces lock contention (the `getCachedValues()` API now requires a lock to be provided).

### DIFF
--- a/src/main/kotlin/org/pkl/lsp/services/DiagnosticsManager.kt
+++ b/src/main/kotlin/org/pkl/lsp/services/DiagnosticsManager.kt
@@ -44,6 +44,8 @@ class DiagnosticsManager(project: Project) : Component(project) {
   private val openFiles: MutableMap<URI, Boolean> = ConcurrentHashMap()
   private val downloadPackageTracker = SimpleModificationTracker()
 
+  private val lock = Object()
+
   override fun initialize(): CompletableFuture<*> {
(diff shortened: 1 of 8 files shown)


src/main/kotlin/org/pkl/lsp/services/DiagnosticsManager.kt
```diff
@@ -96,7 +98,8 @@ class DiagnosticsManager(project: Project) : Component(project) {
 
   fun getDiagnostics(module: PklModule): List<PklDiagnostic> {
     return project.cachedValuesManager.getCachedValue(
-      "DiagnosticsManager.getDiagnostics(${module.uri})"
+      "DiagnosticsManager.getDiagnostics(${module.uri})",
+      lock,
```
bioball, line 102: Previously, this was using the cached values manager itself as the lock, which was the cause for deadlocks to arise (e.g. getDiagnostics -> getLibraryRoots -> getDiagnostics -> deadlock)

stackoverflow (approved):
I wonder if it isn't better to use a `ConcurrentHashMap` instead of a lock, which is design to be lock free.

bioball:
Yeah, maybe a good improvement. They're actually not lock free, though. But, they help with lock contention.

I'll merge this, but we can improve this in a follow-up PR.

PR #93, opened 2025-05-23
Resolve modulepath imports from local file

### LINKED ISSUE #91: support for project/package specific modulepaths
Currently the lsp cannot resolve imports from the `modulepath` (e.g. `import "modulepath:/mymodule/somefile.pkl"`).
It would be nice if a user could somehow provide a local `modulepath` for the lsp to use for the project (or even `package`).

Here is our usecase:
In essence we have a system where our `pkl` files are deployed to locations we do not know beforehand and which are different from the systems they're beeing developed on.
We pass those locations via `--modulepath` and import the files from there. This works for evaluating, but since the `pkl-lsp` does not support `modulepath`s, development is really unpleasant.

We've tried alternatively to instead use `package`s and import the dependencies there via `modulepath`:
```pkl
amends "pkl:Project"

package {
  name = "mypackage"
  // ...
}

dependencies {
  ["myotherpackage"] = import("modulepath:/myotherpackage/PklProject")
}
```

The idea here was that if `pkl package resolve` would write the resulting local path into the `PklProject.deps.json`, then the lsp could resolve package imports (like `import "@mymodule/somefile.pkl"`) and only the `PklProject` file would report _"Cannot resolve import."_.

But it fails because of two reasons:
- the `dependencies` field has a constraint for these locally imported `Project`s `projectFileUri`s to effectively start with `"file://"` ([here](https://github.com/apple/pkl/blob/main/stdlib/Project.pkl#L105))
- the cli tool evaluates `Project`s without `ModulePathResolver` ([here](https://github.com/apple/pkl/blob/main/pkl-core/src/main/java/org/pkl/core/project/Project.java#L85))

The first one can be "worked around" by spoofing the `projectFileUri` field in the imported `Project`s:
```pkl
local spoofedUri: String? = read?("env:MODULE_PATH").ifNonNull((path) -> "file://" + (path as String)
    .split(":")
    .map((e) -> if (!e.endsWith("/")) e + "/" else e)
    .map((e) -> e + module.package.name + "/" + "PklProject")
    .find((e) -> read?("file:" + e) != null))
projectFileUri = if (spoofedUri != null) spoofedUri else super.projectFileUri
```

For the second issue we had to patch the `cli` tool ourselfs, but after that we got the lsp to resolve these `package` dependencies by exporting a `MODULE_PATH` environment variable. 
But obviously this isn't ideal.

If you can point me in the right direction I'd be happy to work on a PR. Or if there is another solution we haven't considered yet, please let me know.  
We would be fine with writing an external reader if the lsp would support it (which might be easier to implement).  
We would also be okay with publishing and fetching packages for development, but when they are deployed we have to use the local files.

sitepark-schaeper:
This aims to resolve #91.

This is a rudementary implementation to figure out, what direction this feature should go.
It looks for a `.pkl-module-path` file in the workspacefolders, reads it, and interpretes ever line that is neither blank nor starts with a `#` (to allow comments) as a `modulepath` entry.

Currently this happens _every time_ an import is resolved. Hence Invalid and non-existent entries are ignored at the moment to not flood the logs. A better way would probably be to use the `CachedValuesManager` with a `dependency` either to the `FsFile` or an lsp event signaling its creation. Alternatively it could load it just once at startup.

Also, is a `.pkl-module-path` file appropriate or should it be more general for lsp settings?

### DIFF
--- a/src/main/kotlin/org/pkl/lsp/ast/PklModuleUri.kt
+++ b/src/main/kotlin/org/pkl/lsp/ast/PklModuleUri.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2024 Apple Inc. and the Pkl project authors. All rights reserved.
+ * Copyright © 2024-2025 Apple Inc. and the Pkl project authors. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,6 +16,7 @@
 package org.pkl.lsp.ast
 
 import io.github.treesitter.jtreesitter.Node
+import java.nio.file.Files
 import java.nio.file.Path
 import org.pkl.lsp.*
 import org.pkl.lsp.FsFile
@@ -121,6 +122,16 @@ class PklModuleUriImpl(project: Project, override val parent: PklNode, override 
             getDependencyRoot(project, targetUriStr, enclosingModule, context)
               ?.resolve(targetUri.fragment)
           vfile?.getModule()?.get()
+        }
+        "modulepath" -> {
+          val path = targetUri.path.trimStart('/')
+          val absolutePath =
+            project.pklProjectManager.findModulePath().map { it.resolve(path) }.first(Files::exists)
+              ?: return null
+          return sourceFile.project.virtualFileManager
+            .get(absolutePath.toUri(), absolutePath)
+            ?.getModule()
+            ?.get()
         }
         // targetUri is a relative URI
         null -> {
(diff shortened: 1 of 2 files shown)


sitepark-schaeper:
> The setting should be part of configuration (see `SettingsManager`)

I've implemented this. Clients now need a way to set this configuration individually per project.

> [...] `import "/foo/bar.pkl"` should look within modulepath

I believe the cli would have to support this first, right?

As far as I can tell resources (from the `read` instruction) are not resolved at all yet and I believe that feature would be beyond the scope of this PR.

PR #102, opened 2025-08-29
Check for NonNull in diagnostics and type checks

stackoverflow:
Because we only check subtype and some constraints relationship, a wrong warning would show
when comparing things to `NonNull`.
```pkl
myOptionalThing: String?

thingsThatAre: Listing<String> = new {
  // this was showing: Expression is always true
  when(myOptionalThing is NonNull) {
    myOptionalThing
  }
}
```

Even though this is not idiomatic Pkl (one should use `??` or compare something to `null` directly) we shouldn't show a wrong warning.

### DIFF
--- a/src/main/kotlin/org/pkl/lsp/analyzers/ExprAnalyzer.kt
+++ b/src/main/kotlin/org/pkl/lsp/analyzers/ExprAnalyzer.kt
@@ -36,7 +36,7 @@ class ExprAnalyzer(project: Project) : Analyzer(project) {
         if (testedType.hasConstraints) return false
         // can't use subtype relationship to check NonNull
         if (testedType.isNonNull(base)) {
-          if (!exprType.isNullable(base, context)) {
+          if (!exprType.isNullable(base, context) || exprType.isNonNull(base)) {
             diagnosticsHolder.addWarning(node, ErrorMessages.create("expressionIsAlwaysTrue"))
           }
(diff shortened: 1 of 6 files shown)


src/main/kotlin/org/pkl/lsp/analyzers/ExprAnalyzer.kt
```diff
@@ -34,10 +34,17 @@ class ExprAnalyzer(project: Project) : Analyzer(project) {
         val exprType = node.expr.computeExprType(base, mapOf(), context)
         val testedType = node.type.toType(base, mapOf(), context)
         if (testedType.hasConstraints) return false
-        if (exprType != Type.Unknown && exprType.isSubtypeOf(testedType, base, context)) {
-          diagnosticsHolder.addWarning(node, ErrorMessages.create("expressionIsAlwaysTrue"))
-        } else if (!testedType.isSubtypeOf(exprType, base, context)) {
-          diagnosticsHolder.addWarning(node, ErrorMessages.create("expressionIsAlwaysFalse"))
+        // can't use subtype relationship to check NonNull
+        if (testedType.isNonNull(base)) {
+          if (!exprType.isNullable(base, context)) {
```
HT154, line 39: What happens here if exprType is itself `NonNull`?

HT154 (approved):

PR #104, opened 2025-09-23
Add configurable link to pkl doc in hover

gordonbondon:
Screenshot 2025-09-23 at 13 46 53
Screenshot 2025-09-23 at 13 48 29

### DIFF
--- a/src/main/kotlin/org/pkl/lsp/ast/Extensions.kt
+++ b/src/main/kotlin/org/pkl/lsp/ast/Extensions.kt
@@ -674,6 +674,11 @@ fun PklNode.getDocumentationUrl(): String? {
       isInStdlib -> {
         module.moduleName?.removePrefix("pkl.") ?: return null
       }
+      moduleFileUri.startsWith("jar:") -> {
+        // Extract the path after the zip file reference
+        val afterJar = moduleFileUri.substringAfter("!/")
+        afterJar.removeSuffix(".pkl")
+      }
       else -> {
         // Fallback: use the module name
         module.moduleName?.removeSuffix(".pkl") ?: return null
--- a/src/main/kotlin/org/pkl/lsp/documentation/toMarkdown.kt
+++ b/src/main/kotlin/org/pkl/lsp/documentation/toMarkdown.kt
@@ -243,6 +243,7 @@ private fun showDocCommentAndModule(node: PklNode?, text: String, context: PklPr
           when (node) {
             is PklModule -> module.moduleName
             is PklProperty -> node.name
+            is PklMethod -> node.methodHeader.identifier?.text?.let { "$it()" } ?: "method"
             is PklMethodHeader -> node.identifier?.text?.let { "$it()" } ?: "method"
             is PklClass -> node.identifier?.text ?: "class"
             is PklTypeAlias -> node.identifier?.text ?: "type"
--- a/src/test/kotlin/org/pkl/lsp/HoverTest.kt
+++ b/src/test/kotlin/org/pkl/lsp/HoverTest.kt
@@ -270,7 +270,7 @@ class HoverTest : LspTestBase() {
     assertThat(hoverText).contains("`String` on pkl-lang.org")
     assertThat(hoverText)
       .contains(
-        "https://pkl-lang.org/package-docs/pkl/${fakeProject.stdlib.version}/base/index.html#String"
+        "https://pkl-lang.org/package-docs/pkl/${fakeProject.stdlib.version}/base/String.html"
       )
   }
 }


src/main/kotlin/org/pkl/lsp/ast/Extensions.kt
```diff
@@ -626,3 +626,102 @@ fun Appendable.renderParameterList(
...
+    when {
+      packageDep == null && file.pklAuthority == Origin.STDLIB.name.lowercase() ->
+        DEFAULT_PKL_AUTHORITY
+      packageDep != null -> packageDep.packageUri.authority
+      else -> return null
+    }
+
+  val packagePath =
+    when {
+      packageDep == null && file.pklAuthority == Origin.STDLIB.name.lowercase() -> {
```
bioball, line 646:
```suggestion
      isInStdlib -> {
```

src/main/kotlin/org/pkl/lsp/ast/Extensions.kt
```diff
@@ -626,3 +626,102 @@ fun Appendable.renderParameterList(
...
+        "pkl"
+      }
+      packageDep != null -> {
+        val packageUri = packageDep.packageUri
+        val authority = packageUri.authority
+
+        authority + "/" + packageUri.path.substringBeforeLast('@').removePrefix("/")
```
bioball, line 653: Make `org.pkl.lsp.packages.dto.PackageUri#basePath` public, and use that instead of `packageUri.path.substringBeforeLast('@')`

src/main/kotlin/org/pkl/lsp/documentation/toMarkdown.kt
```diff
@@ -229,6 +229,34 @@ private fun showDocCommentAndModule(node: PklNode?, text: String, context: PklPr
...
+            java.net.URI(docUrl).host
+          } catch (e: Exception) {
+            "documentation"
+          }
+
+        val linkText =
+          when (node) {
+            is PklModule -> module.moduleName
+            is PklProperty -> node.name
+            is PklMethod -> node.methodHeader.identifier?.text?.let { "$it()" } ?: "method"
```
bioball, line 246: We already have `PklMethodHeader` here, `PklMethod` is the whole method include the method body

```suggestion
```

src/main/kotlin/org/pkl/lsp/documentation/toMarkdown.kt
```diff
@@ -229,6 +229,34 @@ private fun showDocCommentAndModule(node: PklNode?, text: String, context: PklPr
...
       appendLine()
       append("in [${module.moduleName}](${module.getLocationUri(forDocs = true)})")
+
+      val docUrl = node?.getDocumentationUrl()
+      if (docUrl != null) {
+        val domain =
+          try {
+            java.net.URI(docUrl).host
+          } catch (e: Exception) {
+            "documentation"
```
bioball, line 239: This would be a misconfiguration error if this happens; seems like we should catch this at configuration time.

bioball (changes requested):

src/main/kotlin/org/pkl/lsp/ast/Extensions.kt
```diff
@@ -626,3 +626,102 @@ fun Appendable.renderParameterList(
...
+  // Extract the relative module path within the package
+  val moduleFileUri = file.uri.toString()
+  val modulePath =
+    when {
+      // Handle jar: URLs from packages
+      moduleFileUri.startsWith("jar:") -> {
+        // Extract the path after the zip file reference
+        val afterJar = moduleFileUri.substringAfter("!/")
+        afterJar.removeSuffix(".pkl")
+      }
```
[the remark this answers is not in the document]
gordonbondon: With this change it gets me:
```
"https://pkl-lang.org/package-docs/pkg.pkl-lang.org/pkl-k8s/k8s/1.2.1/k8s.k8sSchema/index.html"
```
instead of expected:
```
"https://pkl-lang.org/package-docs/pkg.pkl-lang.org/pkl-k8s/k8s/1.2.1/k8sSchema/index.html"
```
So `k8s.k8sSchema` instead of `k8sSchema` for module name.

src/main/kotlin/org/pkl/lsp/documentation/toMarkdown.kt
gordonbondon: With this removed link text for me instead of `hasUniquePortNames()` defaulted to a module name `k8s.K8sObject` so PklMethodHeader is not matching, in test case:
```
local result = new K8sObject {}.hasUnique<caret>PortNames()
```

src/main/kotlin/org/pkl/lsp/PklClientOptions.kt
```diff
@@ -24,6 +26,17 @@ data class PklClientOptions(
...
+
+  /**
+   * Map of package server authorities to their documentation URL patterns. The URL pattern should
+   * contain placeholders: {packagePath}, {version}, {modulePath}, {path} Example:
+   * "https://pkl-lang.org/package-docs/{packagePath}/{version}/{modulePath}/{path}"
+   */
+  val packageDocumentationUrls: Map<String, String> =
+    mapOf(
+      DEFAULT_PKL_AUTHORITY to
+        "https://pkl-lang.org/package-docs/{packagePath}/{version}/{modulePath}/{path}"
```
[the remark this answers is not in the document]
gordonbondon: For future improvement, how about having a url shema for docs similar to source code schema in package description? https://pkl-lang.org/package-docs/pkl/current/Project/Package.html#sourceCodeUrlScheme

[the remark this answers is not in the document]
bioball: godoc is a central docsite (all public go modules are available on godoc).
Currently, we don't have the concept of a central pkldoc site, but this is something that we've talked about hosting in the past.
If we have that, then we only need one setting for "this is where the docsite lives".

> How about we have a config with package/module prefixes mapped to doc site url prefix?

For the `pkg.pkl-lang.org` stuff, we kind of need a "negative match" on packages that start with `pkg.pkl-lang.org/github.com`, rather than a prefix match.

> For future improvement, how about having a url shema for docs similar to source code schema in package description? https://pkl-lang.org/package-docs/pkl/current/Project/Package.html#sourceCodeUrlScheme

The package metadata actually does have a field for documentation: https://pkl-lang.org/package-docs/pkl/current/Project/Package.html#documentation. This field will appear in the package metadata if set.
Perhaps, for now, we can just use that field in lieu of first checking a docsite.

PR #107, opened 2025-11-04
Add support for formatting

bioball:
This uses the Pkl formatter library to format Pkl code.

It defines a new configuration setting calling `pkl.formatter.grammarVersion`, which language clients should define, or it falls back to grammar version 2.


gradle/libs.versions.toml
HT154: We should update both of these as well, right?

bioball: that's in this PR: https://github.com/apple/pkl-lsp/pull/108

HT154 (approved):

PR #141, opened 2026-01-17
Support modulepath imports

### LINKED ISSUE #91: support for project/package specific modulepaths

DrWursterich:
Closes #91

I saw #93 and #120 and adopted both approaches, the `SettingsManager` and the `EvaluatorSettings`. I decided to priorize the `SettingsManager`'s `modulepath` when resolving since that could offer a way to "override" a projects `modulepath` locally.

image

I also added completion for `import`, `amends` ect.  

image

image

---

### Addressing [the feedback on #93](https://github.com/apple/pkl-lsp/pull/93#issuecomment-2905761737)

> The setting should be part of configuration (see `SettingsManager`). On the client side, this should be configured differently depending on the editor (in VSCode, it should be defined in the extension's package.json)

Should be solved, right? We do use the `SettingsManager` here, and everything else would have to be done in [pkl-vscode](https://github.com/apple/pkl-vscode). While looking into the now-botched #93 I found that @sitepark-schaeper has [a version of that](https://github.com/apple/pkl-vscode/compare/main...sitepark-schaeper:pkl-vscode:main), but I don't know if that works or not. It was trivial for neovim.

> When resolving relative paths, we should determine whether that file exists within the modulepath or not. If so, relative paths should be resolved according to the modulepath. For example, `import "/foo/bar.pkl"` should look within modulepath.

This is covered specifically by the test cases [`resolve relative field in modulepath`](https://github.com/apple/pkl-lsp/compare/main...DrWursterich:pkl-lsp:main?expand=1#diff-3f845d525e8920a6d14d6dbf2a20b0048ede37dad5c22b8d07251c994203510aR270) and [`resolve modulepath field in modulepath`](https://github.com/apple/pkl-lsp/compare/main...DrWursterich:pkl-lsp:main?expand=1#diff-3f845d525e8920a6d14d6dbf2a20b0048ede37dad5c22b8d07251c994203510aR313).

> Resolving a modulepath resource should look at each jar/directory (as determined by the order of jars/dirs) and return the first file that exists. This follows the same rules as classpath resources in Java.

The implementation does exactly that. It iterates through all `modulepath` entries (first from the `SettingsManager`, then from the `EvaluatorSettings`) and the first file that exists is used.
Soley jar entries have not been tested, as I am unsure how to do so. If you need me to do that as well please help me setup a test to implement against.

---

Please let me know if there’s anything else I can improve or clarify to help move this feature forward. I’m happy to iterate further.


DrWursterich:
Hi @bioball - sorry to tag you directly. I'd really appreciate a review when you have a chance. Thanks!

HT154:
Sorry for the delay on this! We're a bit of a skeleton crew at the moment. We haven't forgotten this and will take a look when we can.

stackoverflow (commented):
Without jar file support I'm not sure how useful that would be.
Most of the use case for modulepath is for direct jvm usage in Java/Kotlin/etc. where users load pkl files from jars.

DrWursterich:
Thanks for the review!
I agree, jar support is important here. However, even without it, there is still a need for this feature, as I am now the third person attempting to add it.

I’ve started looking into implementing jar support for modulepaths and will report back - likely toward the end of the week.

DrWursterich:
Okay, I've added the archive support for modulepath imports now.

One open question is the uri format exposed to the client. Apparently, there is no clear standard.  
Internally, the lsp has to use `jar:file:///some/archive.jar!/file/inside.pkl` since it runs on the jvm, but client support for that seems very limited, if present at all. Other language servers sometimes use `zipfile:///some/archive.jar::file/inside.pkl`, which appears to be more widely supported, so that's what I went with.

src/main/kotlin/org/pkl/lsp/ast/Extensions.kt
```diff
@@ -417,6 +417,7 @@ val VirtualFile.lspUri
     return when (this) {
       is StdlibFile -> URI("pkl-lsp://stdlib/${name}.pkl")
       is FsFile -> uri
+      is JarFile -> URI("zip${uri.schemeSpecificPart.replaceFirst("!/", "::")}")
```
HT154: I don't think this change is necessary. The `else` case should be sufficient to correctly encode the `pkl-lsp:` URI, which clients should already support opening.

src/main/kotlin/org/pkl/lsp/services/ModulepathResolver.kt
```diff
@@ -0,0 +1,68 @@
...
+    val modulepath = context.metadata.evaluatorSettings?.modulePath ?: return emptyList()
+    return modulepath.map { this.normalizeArchivePath(context.projectDir.resolve(it)) }
+  }
+
+  private fun all(context: PklProject?): List<Path> =
+    fromSettings() + context?.let(::fromProject).orEmpty()
+
+  private fun normalizeArchivePath(path: Path): Path {
+    val path = path.normalize()
+    if (Files.isRegularFile(path) && archivePathMatcher.matches(path)) {
```
HT154: The Pkl core implementation understands that a module path element is a zip/jar if:
* It is a regular file ✅
* It has the extension `.zip` or `.jar` ✅
* The file starts with a zip header ❌
* The file starts with an executable jar header ❌

It would probably be good to mirror [the core implementation here](https://github.com/apple/pkl/blob/eb3891b21f375081d2553e3a8f11e6bf7856d6e6/pkl-core/src/main/java/org/pkl/core/module/ModulePathResolver.java#L168) for the best user experience.

src/main/kotlin/org/pkl/lsp/VirtualFileManager.kt
```diff
@@ -54,6 +54,11 @@ class VirtualFileManager(project: Project) : Component(project) {
           ensureJarFileSystem(effectiveUri)
           JarFile(path ?: Path.of(effectiveUri), effectiveUri, project)
         }
+        "zipfile" -> {
+          val jarUri = URI("jar:file:${effectiveUri.schemeSpecificPart.replaceFirst("::", "!/")}")
+          ensureJarFileSystem(jarUri)
+          JarFile(path ?: Path.of(jarUri), jarUri, project)
+        }
```
HT154: Likewise, if you rely on the existing `pkl-lsp:` scheme for URIs, this shouldn't be necessary.

src/test/kotlin/org/pkl/lsp/CompletionFeatureTest.kt
```diff
@@ -24,22 +24,49 @@ class CompletionFeatureTest : LspTestBase() {
...
   fun `complete import expression`() {
     createPklFile("res = import(\"<caret>\")")
     val completions = getCompletions()
     assertThat(completions.map { it.label })
-      .isEqualTo(listOf("pkl:", "file:///", "https://", "package://", "main.pkl"))
+      .isEqualTo(listOf("pkl:", "file:///", "https://", "package://", "modulepath:/", "main.pkl"))
+  }
+
+  @Test
+  fun `complete modulepath import`() {
```
HT154, line 48: Missing coverage: testing that matches earlier in then modulepath take precedence over later matches.

src/test/kotlin/org/pkl/lsp/GoToDefinitionTest.kt
```diff
@@ -218,4 +218,155 @@ class GoToDefinitionTest : LspTestBase() {
...
+    val resolved = goToDefinition()
+    assertThat(resolved).hasSize(1)
+    assertThat(resolved[0]).isInstanceOf(PklClassProperty::class.java)
+    assertThat((resolved[0] as PklClassProperty).name).isEqualTo("field")
+    val resolvedFile = resolved.first().containingFile
+    assertThat(resolvedFile.uri.path).endsWith("/lib/Target.pkl")
+  }
+
+  @Test
+  fun `resolve relative field in modulepath`() {
```
HT154, line 290: Missing coverage: relative resolution across multiple modulepath elements, e.g.
```
$ tree x # with zip and text file contents added
x
├── y
│   └── foo.pkl  | bar = import("./bar.pkl") // relative import to a different modulepath element
├── z
│   └── bar.pkl  | baz = 0
├── test.pkl     | foo = import("modulepath:/foo.pkl")
└── z.zip
    └── bar.pkl  | baz = 1

$ pkl eval x/test.pkl --module-path x/y:x/z:x/z.zip # directory before zip
foo {
  bar {
    baz = 0
  }
}

$ pkl eval x/test.pkl --module-path x/y:x/z.zip:x/z # zip before directory
foo {
  bar {
    baz = 1
  }
}
```

This case is not covered correctly by the implementation: when a module on the modulepath imports relatively, the import should be resolved on the module path. The current implementation only imports relative to the effective real `file:` URI of importing module.

src/main/kotlin/org/pkl/lsp/services/ModulepathResolver.kt
```diff
@@ -0,0 +1,68 @@
...
+    return project.settingsManager.settings.modulepath.map(this::normalizeArchivePath)
+  }
+
+  private fun fromProject(context: PklProject): List<Path> {
+    val modulepath = context.metadata.evaluatorSettings?.modulePath ?: return emptyList()
+    return modulepath.map { this.normalizeArchivePath(context.projectDir.resolve(it)) }
+  }
+
+  private fun all(context: PklProject?): List<Path> =
+    fromSettings() + context?.let(::fromProject).orEmpty()
```
HT154: Looking for input from other reviewers on what the behavior should be here. In most cases in Pkl, configuring something at one level completely overrides lower-precedence configuration levels, e.g. setting an evaluator setting in a PklProject overrides (not merges) the setting from a [settings file](https://pkl-lang.org/main/current/pkl-cli/index.html#settings-file); likewise, setting an evaluator setting on the CLI prevents loading it from both PklProject and a settings file.

I'm leaning towards: if the LSP gets a ModulePath from a PklProject, then that should completely override the module path as set in the LSP settings.

HT154 (changes requested):
Thanks for contributing this! Sorry it took so long to review. The implementation largely seems fine, but there are some gaps that prevent merging this:
* Use the existing `pkl-lsp:` URI scheme instead of introducing `zipfile:`
* Missing support for cross-modulepath-element relative importing
* An open question around precedence/merging of project- vs. settings-defined modulepath

I'd like one of the other core maintainers to weigh in on that last bit, but that will take a couple weeks at least. If you'd like to address the other bits in the meantime, I can take another look to hopefully streamline landing this!

DrWursterich:
Thank you very much for the detailed feedback!

I have started working on it and have two follow-up questions:

---

> Use the existing `pkl-lsp:` URI scheme instead of introducing `zipfile:`

Exclusively using `pkl-lsp:` means, that every single client would need its own plugin/extension/... to handle `pkl-lsp` uris. I'm sure the vscode plugin has this ability but not everyone uses vscode. In my case it's neovim and there is no pkl plugin for that - as it's not needed. All I had to do was to configure and enable the lsp
```lua
vim.lsp.config('pkl-lsp', {
    filetypes = { 'pkl' },
    cmd = { '/usr/local/bin/pkl-lsp' },
    -- ...
})
vim.lsp.enable('pkl-lsp')
```
The same applies for at least Emacs, Helix, Zed and Sublime.

Perhaps this could be addressed by deciding this on a per-client basis. Either via client-settings or by identifying the client in the `initialize` request.
But if you do prefer, I can also just remove the `zipfile:` feature.

---

> Missing support for cross-modulepath-element relative importing

I really appreciate the detailed scenario you provided. However, I'm not convinced it should behave like that. On my end (Pkl 0.32.0-dev+3eda3b19) the provided scenario did not resolve the cross-modulepath-import:
```
Cannot find module `modulepath:/bar`.

1 | bar = import("./bar")
          ^^^^^^^^^^^^^^^
at foo#bar (modulepath:/foo.pkl)
```
This aligns with my expectations. If `import "./bar.pkl"` looked into `modulepath` entries, it would be very easy to accidentally have libraries "override" files in other libraries.

Please let me know if I am still missing something here!

HT154:
> Exclusively using pkl-lsp: means, that every single client would need its own plugin/extension/... to handle pkl-lsp uris. I'm sure the vscode plugin has this ability but not everyone uses vscode. In my case it's neovim and there is no pkl plugin for that - as it's not needed. All I had to do was to configure and enable the lsp

It's already necessary for clients to handle the `pkl-lsp:` scheme to correctly support opening the stdlib and package-based modules. If we do revise this in the future, it would be good to do everything at the same time. In the meantime, I think consistency with existing functionality is preferable.

> I really appreciate the detailed scenario you provided. However, I'm not convinced it should behave like that. On my end (Pkl 0.32.0-dev+3eda3b19) the provided scenario did not resolve the cross-modulepath-import.

Sorry, this was an error in transcription. `import("./bar")` should have been `import("./bar.pkl")`. Here's the exact [`x` directory](https://github.com/user-attachments/files/25977510/x.zip) I'm testing with. You should be able to observe this behavior.

> This aligns with my expectations. If `import "./bar.pkl"` looked into `modulepath` entries, it would be very easy to accidentally have libraries "override" files in other libraries.

`modulepath:` is designed to work very similarly to the Java class path. As such, the ability to override module in other libraries is _exactly_ the expected behavior!

DrWursterich:
Sorry for the embarrassing mistake, and thanks for pointing it out. I can now confirm the behaviour you've described. I have updated the implementation and written tests for this scenario.
I've also removed the `zipfile:` scheme support.

Please let me know if this looks correct and if there's anything more to address.

src/main/kotlin/org/pkl/lsp/ast/PklModuleUri.kt
```diff
@@ -229,11 +229,20 @@ class PklModuleUriImpl(project: Project, override val parent: PklNode, override
...
-        targetPath.startsWith(".../") -> findTripleDotPathOnFileSystem(sourceFile, targetPath)
-        targetPath.startsWith("/") -> findByAbsolutePath(sourceFile, targetPath)
-        else -> sourceFile.parent()?.resolve(targetPath)
+        targetPath.startsWith(".../") ->
+          findTripleDotPathOnFileSystem(sourceFile, targetPath)
+            ?: project.modulepathResolver.resolve(targetPath, context)
+        targetPath.startsWith("/") -> findByAbsolutePath(sourceFile, targetPath)?.getModule()?.get()
+        else ->
+          sourceFile.parent()?.resolve(targetPath)?.getModule()?.get()
+            ?: project.modulepathResolver.resolve(targetPath, context)
```
HT154: The modulepath resolver (and only the modulepath resolver) should be used when `sourceFile`'s module has a `modulepath:` URI. This should probably be a distinct case in the `when` up near line 127 instead of a modification to `findOnFileSystem`.

It would be good to add some tests here to assert:
* Imports in `file:` sources can't unintentionally import modules on the `modulepath:` by their relative path
* Imports in `modulepath:` sources can't unintentionally import modules relatively (on the filesystem) that are outside the modulepath

DrWursterich: > Imports in `file:` sources can't unintentionally import modules on the `modulepath:` by their relative path

Ohh, I thought this _should_ be possible, which is why I went to `findOnFileSystem`..

I've played around more and think I _finally_ got it now; Makes a lot more sense this way :D

src/main/kotlin/org/pkl/lsp/ast/PklModuleUri.kt
```diff
@@ -140,6 +137,8 @@ class PklModuleUriImpl(project: Project, override val parent: PklNode, override
                 root.resolve(resolvedTargetUri)?.getModule()?.get()
               } else null
             }
+            sourceFile is FsFile && sourceFile.isOnModulepath ->
```
HT154, line 141: `JarFile` instances may be on the module path as well, do those need to also be specifically tracked?

src/main/kotlin/org/pkl/lsp/services/ModulepathResolver.kt
```diff
@@ -0,0 +1,97 @@
...
+    return modulepath
+      .asSequence()
+      .map { it.resolve(path).normalize() }
+      .firstOrNull(Files::exists)
+      ?.let { project.virtualFileManager.get(URI.create("modulepath:${it.toUri()}"), it) }
+      ?.getModule()
+      ?.get()
+  }
+
+  fun resolveDirect(path: String, context: PklProject?): PklModule? =
```
HT154:
```suggestion
  fun resolveAbsolute(path: String, context: PklProject?): PklModule? =
```

src/main/kotlin/org/pkl/lsp/ast/PklModuleUri.kt
DrWursterich: the source file of an import has to be a file containing source code, no?

HT154: A file within a zip/jar on the modulepath can contain relative imports that need to be resolved against the modulepath, right? Or am I missing something here?

DrWursterich: Well, yes, a `JarFile` could also have an `isOnModulepath` property.  
What I'm saying is that we only need to know whether a file is on the modulepath if it includes a relative import, to determine whether that's allowed to potentially resolve onto another modulepath entry.  
If a file has an import that resolves to a file on the modulepath (within an archive or not), then the `ModulepathResolver` will cause the `VirtualFileManager` to set that flag regardless of any parent archives or directories.

If you want to, I can add this property to all `JarFile`s and `FsFile` directories that technically are on the modulepath, but I currently don't see a reason for it.

src/main/kotlin/org/pkl/lsp/ast/Extensions.kt
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2024-2025 Apple Inc. and the Pkl project authors. All rights reserved.
+ * Copyright © 2024-2026 Apple Inc. and the Pkl project authors. All rights reserved.
```
stackoverflow: You can revert this.

src/main/kotlin/org/pkl/lsp/ast/PklModuleUri.kt
stackoverflow: We should add `isOnModulePath` to JarFile also. Even if the code is correct as it is (I'm not 100% sure if there's some edge case we're missing), it's more principled to have it there, and communicates better the intent.

src/main/kotlin/org/pkl/lsp/services/ModulepathResolver.kt
stackoverflow: Agree with Jen here. We should override, not merge (same logic as Pkl).

src/main/kotlin/org/pkl/lsp/completion/ModuleUriCompletionProvider.kt
```diff
@@ -183,6 +185,20 @@ class ModuleUriCompletionProvider(project: Project, private val packageUriOnly:
           }
         }
       }
+      targetUri.startsWith(MODULEPATH_SCHEME) -> {
+        val context = sourceModule.virtualFile.pklProject
+        val roots = project.modulepathResolver.paths(context)
+        completeHierarchicalUri(
+          roots,
+          ".",
+          targetUri.substring(12),
```
stackoverflow:
```suggestion
          targetUri.substring(MODULEPATH_SCHEME.length),
```

src/main/kotlin/org/pkl/lsp/services/ModulepathResolver.kt
```diff
@@ -0,0 +1,97 @@
...
+  private fun fromProject(context: PklProject): List<Path> {
+    val modulepath = context.metadata.evaluatorSettings?.modulePath ?: return emptyList()
+    return modulepath.map { this.normalizeArchivePath(context.projectDir.resolve(it)) }
+  }
+
+  private fun all(context: PklProject?): List<Path> =
+    fromSettings() + context?.let(::fromProject).orEmpty()
+
+  private fun normalizeArchivePath(path: Path): Path {
+    val path = path.normalize().toAbsolutePath()
```
stackoverflow:
```suggestion
    val normalizedPath = path.normalize().toAbsolutePath()
```
nit: shadowing

DrWursterich:
I think I’ve addressed all the feedback. Please let me know if it looks okay now.

mbUSC:
+1 this will be great to have

src/main/kotlin/org/pkl/lsp/services/ModulepathResolver.kt
```diff
@@ -0,0 +1,94 @@
...
+class ModulepathResolver(project: Project) : Component(project) {
+
+  private val archivePathMatcher: PathMatcher by lazy {
+    FileSystems.getDefault().getPathMatcher("glob:**/*.{zip,jar}")
+  }
+
+  private fun modulepaths(context: PklProject?): List<Path> {
+    val paths =
+      context?.metadata?.evaluatorSettings?.modulePath?.map(context.projectDir::resolve)
+        ?: project.settingsManager.settings.modulepath
```
bioball, line 38: Two comments:

1. This logic is actually the opposite of your PR description (you say settings has precedence over PklProject, but this is putting PklProject first)
2. I actually think it makes sense to merge the directly configured modulepath together with `PklProject`. The LSP doesn't know how the Pkl evaluator is configured; it can either be configured using PklProject or something else. It's better if it's a best-effort thing and resolve them if users tell the LSP about it in any way (CC @HT154 @stackoverflow)

src/main/kotlin/org/pkl/lsp/services/ModulepathResolver.kt
```diff
@@ -0,0 +1,94 @@
...
+      val uri = URI.create("jar:${normalizedPath.toUri()}!/")
+      ensureJarFileSystem(uri)
+      return Paths.get(uri)
+    }
+    return normalizedPath
+  }
+
+  private fun isArchive(path: Path): Boolean {
+    if (!Files.isRegularFile(path)) return false
+    if (archivePathMatcher.matches(path)) return true
```
bioball:
```suggestion
    if (path.extension == "zip" || path.extension == "jar") return true
```

src/main/kotlin/org/pkl/lsp/services/ModulepathResolver.kt
```diff
@@ -0,0 +1,94 @@
...
+    val execHeader =
+      "#!/bin/sh\n      exec java  -jar $0 \"$@\"".toByteArray(StandardCharsets.UTF_8)
```
bioball: Create a `companion object` and move `execHeader` into there

src/test/kotlin/org/pkl/lsp/GoToDefinitionTest.kt
```diff
@@ -218,4 +219,296 @@ class GoToDefinitionTest : LspTestBase() {
...
+          baz<caret> = 2
+        }
+      }
+    """
+        .trimIndent(),
+    )
+    var resolved = goToDefinition().single()
+    assertThat(resolved).isInstanceOf(PklClassProperty::class.java)
+    var uri = resolved.enclosingModule!!.uri
+    var path = Paths.get(uri.path).normalize().toString()
```
bioball:
```suggestion
    val resolved = goToDefinition().single()
    assertThat(resolved).isInstanceOf(PklClassProperty::class.java)
    val uri = resolved.enclosingModule!!.uri
    val path = Paths.get(uri.path).normalize().toString()
```

src/main/kotlin/org/pkl/lsp/VirtualFileManager.kt
```diff
@@ -54,6 +54,7 @@ class VirtualFileManager(project: Project) : Component(project) {
           ensureJarFileSystem(effectiveUri)
           JarFile(path ?: Path.of(effectiveUri), effectiveUri, project)
         }
+        "modulepath" -> FsFile(path ?: Path.of(effectiveUri), this.project, isOnModulepath = true)
```
bioball: If the `Path` is a ZipPath, we should already know to return a `JarFile` here.

If not, we should consult the configured modulepath to determine if we should be creating an `FsFile` or `JarFile`. This suggestion depends on changing the signatures of the various `ModulepathResolver#resolve` methods to return `VirtualFile?`

```suggestion
          if (path?.toUri()?.scheme == "jar") JarFile(path, uri, this.project, isOnModulepath = true)
          else project.modulepathResolver.resolveAbsolute(uri.path, null) ?: FsFile(
            path ?: Path.of(effectiveUri),
            this.project,
            isOnModulepath = true
          )
```

src/main/kotlin/org/pkl/lsp/services/ModulepathResolver.kt
```diff
@@ -0,0 +1,94 @@
...
+          buffer.contentEquals(execHeader)
+      }
+    } catch (_: Exception) {
+      false
+    }
+  }
+
+  private fun resolve(path: String, modulepath: List<Path>): PklModule? {
+    return modulepath
+      .asSequence()
```
bioball:
```suggestion
```

src/main/kotlin/org/pkl/lsp/services/SettingsManager.kt
```diff
@@ -78,6 +80,24 @@ class SettingsManager(project: Project) : Component(project) {
...
+    if (value.isJsonNull) return emptyList()
+    if (!value.isJsonArray) {
+      logger.warn("Got non-array value for configuration: pkl.modulepath. Value: $value")
+      return emptyList()
+    }
+    return buildList {
+      for (path in value.asJsonArray) {
+        val decodedPath = decodeString(path, "pkl.modulepath")
+        if (decodedPath != null) {
+          val entry = Path.of(decodedPath)
```
bioball: Need to resolve relative paths off of the user's workspace. Otherwise it gets resolve off of where PWD, which is probably incorrect.

`SettingsManager#initialize` can receive `folders: List<Path>?` as an argument, and use this to look for relative paths (see how `Stdlib` and `PklProjectManager` do this). It'll also need to subscribe to workspace folder events, too.

src/main/kotlin/org/pkl/lsp/services/SettingsManager.kt
```diff
@@ -78,6 +80,24 @@ class SettingsManager(project: Project) : Component(project) {
...
 
+  private fun resolveModulepath(value: JsonElement): List<Path> {
+    if (value.isJsonNull) return emptyList()
+    if (!value.isJsonArray) {
+      logger.warn("Got non-array value for configuration: pkl.modulepath. Value: $value")
+      return emptyList()
+    }
+    return buildList {
+      for (path in value.asJsonArray) {
+        val decodedPath = decodeString(path, "pkl.modulepath")
```
bioball:
```suggestion
        val decodedPath = decodeString(path, "pkl.modulepath[]")
```

src/main/kotlin/org/pkl/lsp/VirtualFile.kt
```diff
@@ -186,11 +190,9 @@ class FsFile(override val path: Path, override val project: Project) : BaseFile(
     get() = path.isDirectory()
 
   override val children: List<VirtualFile>?
-    get() =
-      if (!isDirectory) null
-      else path.listDirectoryEntries().mapNotNull { project.virtualFileManager.get(it) }
+    get() = if (!isDirectory) null else path.listDirectoryEntries().mapNotNull(::getFile)
```
bioball:
```suggestion
    get() = when {
      !isDirectory -> null
      isOnModulepath -> project.modulepathResolver.listChildren(path.toString(), null)
      else -> path.listDirectoryEntries().mapNotNull(::getFile)
    }
```

bioball: And add this method to `ModulePathResolver`:

```kt
  fun listChildren(path: String, context: PklProject?): List<VirtualFile> {
    return modulepaths(context)
      .flatMap { it.resolve(path.trimStart('/')).normalize().listDirectoryEntries() }
      // in case of duplicate path, first entry wins
      .distinctBy { it.toString() }
      .mapNotNull(::getFile)
  }
```

src/main/kotlin/org/pkl/lsp/services/ModulepathResolver.kt
```diff
@@ -0,0 +1,94 @@
...
+      .map { it.resolve(path).normalize() }
+      .firstOrNull(Files::exists)
+      ?.let(::getFile)
+      ?.getModule()
+      ?.get()
+  }
+
+  private fun getFile(path: Path): VirtualFile? =
+    project.virtualFileManager.get(URI.create("modulepath:${path.toUri()}"), path)
```
bioball:
```suggestion
    project.virtualFileManager.get(URI.create("modulepath:$path"), path)
```

src/main/kotlin/org/pkl/lsp/VirtualFile.kt
```diff
@@ -203,6 +205,11 @@ class FsFile(override val path: Path, override val project: Project) : BaseFile(
 
   override fun doReadContents(): String = path.readText()
 
+  private fun getFile(path: Path): VirtualFile? =
+    if (isOnModulepath)
+      project.virtualFileManager.get(URI.create("modulepath:${path.toUri()}"), path)
```
bioball: Modulepath URIs have the form: `modulepath:/foo/bar.pkl`

This would have created `modulepath:jar:file:///etc!/foo/bar.pkl"`

```suggestion
      project.virtualFileManager.get(URI.create("modulepath:$path"), path)
```

src/main/kotlin/org/pkl/lsp/VirtualFile.kt
```diff
@@ -324,6 +333,11 @@ class JarFile(override val path: Path, override val uri: URI, override val proje
     }!!
 
   override fun canModify(): Boolean = false
+
+  private fun getFile(path: Path): VirtualFile? =
+    if (isOnModulepath)
+      project.virtualFileManager.get(URI.create("modulepath:${path.toUri()}"), path)
```
bioball:
```suggestion
      project.virtualFileManager.get(URI.create("modulepath:$path"), path)
```

src/main/kotlin/org/pkl/lsp/VirtualFile.kt
```diff
@@ -289,9 +300,7 @@ class JarFile(override val path: Path, override val uri: URI, override val proje
     get() = path.isDirectory()
 
   override val children: List<VirtualFile>?
-    get() =
-      if (!isDirectory) null
-      else path.listDirectoryEntries().mapNotNull { project.virtualFileManager.get(it) }
+    get() = if (!isDirectory) null else path.listDirectoryEntries().mapNotNull(::getFile)
```
bioball:
```suggestion
    get() = when {
      !isDirectory -> null
      isOnModulepath -> project.modulepathResolver.listChildren(path.toString(), null)
      else -> path.listDirectoryEntries().mapNotNull(::getFile)
    }
```

src/test/kotlin/org/pkl/lsp/GoToDefinitionTest.kt
```diff
@@ -218,4 +219,296 @@ class GoToDefinitionTest : LspTestBase() {
...
+        bar {
+          baz<caret> = 2
+        }
+      }
+    """
+        .trimIndent(),
+    )
+    val resolved = goToDefinition().single()
+    assertThat(resolved).isInstanceOf(PklClassProperty::class.java)
+    var uri = resolved.enclosingModule!!.uri
```
bioball:
```suggestion
    val uri = resolved.enclosingModule!!.uri
```

bioball (changes requested):
Hey, thanks for the contribution! Sorry it took so long to review this; I was on leave for some time.

This mostly looks great! I left some comments; some nits and some blockers.

src/main/kotlin/org/pkl/lsp/services/ModulepathResolver.kt
```diff
@@ -0,0 +1,94 @@
...
+        (buffer.size >= 4 &&
+          buffer.copyOfRange(0, 4).contentEquals(byteArrayOf(0x50, 0x4b, 0x03, 0x04))) ||
+          buffer.contentEquals(execHeader)
+      }
+    } catch (_: Exception) {
+      false
+    }
+  }
+
+  private fun resolve(path: String, modulepath: List<Path>): PklModule? {
```
bioball: These methods should probably just return `VirtualFile?`, which is much more flexible than returning `PklModule?`. Those that want to use this as a module can call `.getModule()`.

src/main/kotlin/org/pkl/lsp/services/ModulepathResolver.kt
DrWursterich: > This logic is actually the opposite of your PR description

yes, it has changed as a result of [this](https://github.com/apple/pkl-lsp/pull/141#discussion_r2927777467).

before i update the logic here again, it would be good to first agree on how this should work. i don’t have a strong preference either way.

src/main/kotlin/org/pkl/lsp/VirtualFile.kt
DrWursterich: I have built this, but without passing the `context` - which `FsFile::children`/`JarFile::children` do not have access to - modulepaths configured in a `PklProject` are not considered, which in turn affects completion.

Also, the implementation had to look slightly different as the `path` has to first be relativized to whatever root the modulepath is in.

src/main/kotlin/org/pkl/lsp/VirtualFile.kt
DrWursterich: is no longer relevant as `getModulepathFile` is now called instead without `URI`.

src/main/kotlin/org/pkl/lsp/VirtualFileManager.kt
DrWursterich: I have implemented this slightly differently (see my message below).

Is it correct, that every `"jar:"` uri results in a `JarFile`? I mean the `"jar:file///some/file.jar!/"` probably should, but another file inside of a jar as well? is that not just a `FsFile` on another filesystem?
I've assumed you meant this too, so this is how the new solution does it.

src/main/kotlin/org/pkl/lsp/services/SettingsManager.kt
DrWursterich: It's not resolved against the pwd but the `context.projectDir` but i get the point.

please take a look at my solution. it's a bit awkward since we do not keep track of the unresolved modulepaths so we have to request settings again when the workspace folders change. Would you want the `SettingsManager` to store those instead?

DrWursterich:
Thank you for the feedback!
I've tried to implement it, but had to change a couple things up:

---

What previously was `virtualFileManager.get(URI.create("modulepath:${path.toUri()}"), path)` is now `virtualFileManager.getOnModulepath(path)`. You've pointed out correctly that the string substitution resulted in `"modulepath:jar:file:///..."` and suggested a more involved implementation for the `VirtualFileManager::create` `"modulepath"` case. I had built that and then realized that all required information is already in the `path` argument and the `uri` only served as a flag to set the `isOnModulepath` property - which still had to be created for every call. I feel that it's more consistent now and hope it aligns with your expectations.

---

modulepaths from a `PklProject` are with the new `FsFile`/`JarFile` `children` implementation currently not considered, which affects completion. I'd like some guidance on how to solve this architectually.

src/main/kotlin/org/pkl/lsp/services/ModulepathResolver.kt
```diff
@@ -0,0 +1,100 @@
+/*
+ * Copyright © 2024-2026 Apple Inc. and the Pkl project authors. All rights reserved.
```
bioball:
```suggestion
 * Copyright © 2026 Apple Inc. and the Pkl project authors. All rights reserved.
```

src/main/kotlin/org/pkl/lsp/VirtualFileManager.kt
bioball: Practically, I don't think it's important to consider nested jars here; Pkl itself doesn't support loading modules from inside a nested jar. So yeah, everything that has scheme `jar:` should just become a `JarFile`.

src/main/kotlin/org/pkl/lsp/services/SettingsManager.kt
bioball: I see you've updated the logic to resolve against workspace folders, which seems correct to me!

bioball (approved):
Thank you for your contribution! Great work!

I have some adjustments to make, but I'll submit them as a follow-up PR so you don't have to deal with more churn.

HT154 (approved):
LGTM! Thanks a ton for pushing this over the line, we really appreciate the contribution.

DrWursterich:
Awesome, thank you so much!

PR #156, opened 2026-04-03
Fix inferred import names not quoting keywords

DrWursterich:
Closes #139

I took the list of keywords from [here](https://github.com/apple/pkl/blob/main/pkl-parser/src/main/java/org/pkl/parser/Lexer.java#L776) and copied the logic for when to quote identifieres from [here](https://github.com/apple/pkl/blob/main/pkl-parser/src/main/java/org/pkl/parser/Lexer.java#L768).

### DIFF
--- a/src/main/kotlin/org/pkl/lsp/ast/Extensions.kt
+++ b/src/main/kotlin/org/pkl/lsp/ast/Extensions.kt
@@ -226,17 +226,17 @@ val PklImport.memberName: String?
         ?.stringConstant
         ?.escapedText()
         ?.let { inferImportPropertyName(it) }
-        ?.let { if (shouldQuoteIdentifier(it)) "`$it`" else it }
+        ?.let { if (isRegularIdentifier(it)) it else "`$it`" }
 
-private fun shouldQuoteIdentifier(identifier: String): Boolean {
-  if (identifier.isEmpty()) return true
-  if (identifier in KEYWORDS) return true
+private fun isRegularIdentifier(identifier: String): Boolean {
+  if (identifier.isEmpty()) return false
+  if (identifier in KEYWORDS) return false
 
   val firstCp = identifier.codePointAt(0)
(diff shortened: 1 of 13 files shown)


src/main/kotlin/org/pkl/lsp/ast/Extensions.kt
```diff
@@ -175,7 +222,24 @@ fun PklClass.hasCommonSubclassWith(other: PklClass, context: PklProject?): Boole
...
+  if (identifier in KEYWORDS) return true
+
+  val firstCp = identifier.codePointAt(0)
+  if (firstCp != '$'.code && firstCp != '_'.code && !Character.isUnicodeIdentifierStart(firstCp))
+    return true
+
+  return !identifier.codePoints().skip(1).allMatch { cp ->
+    cp == '$'.code || Character.isUnicodeIdentifierPart(cp)
+  }
+}
```
HT154: Not a huge deal, but keeping this function as `isRegularIdentifier` and inverting the result at the call site would make future maintenance smoother if we decide to evolve this part of the language.

HT154 (approved):
This looks good to me! Looking for another set of eyes from @bioball or @stackoverflow.

bioball (changes requested):
This isn't quite right; quoted identifiers stand for the same name regardless of whether it's a keyword or not.

For example, this snippet should not highlight an error:

```pkl
import "pkl:yaml"

foo = `yaml`
```

DrWursterich:
> This isn't quite right; quoted identifiers stand for the same name regardless of whether it's a keyword or not.

Sorry, I would consider that a separate issue. I don't see myself addressing that here.

bioball (approved):

bioball:
Pushed up a new commit that resolves names against the un-quoted identifier name

HT154 (approved):
LGTM

PR #170, opened 2026-05-05
Adjustments to modulepath logic

bioball:
The following adjustments are made:

* `isOnModulepath` is removed as an instance field on `FsFile` and `JarFile`. Instead, `ModulePathResolver` keeps track of whether a dir is on the modulepath or not based on context (a directory may be on modulepath in the context of one PklProject, but not another).
* Add logic to resolve relative paths against `modulepath:` if it is configured as part of the modulepath (requires that you set this in configuration settings).
* modulePath declared in `PklProject.evaluatorSettings` is merged with explicitly configured settings

FYI @DrWursterich

### DIFF
--- a/src/main/kotlin/org/pkl/lsp/ast/PklModuleUri.kt
+++ b/src/main/kotlin/org/pkl/lsp/ast/PklModuleUri.kt
@@ -182,7 +182,7 @@ class PklModuleUriImpl(project: Project, override val parent: PklNode, override 
         "file" -> {
           val listChildren = { it: VirtualFile -> it.children ?: emptyList() }
           val targetPath = targetUri.path ?: return null
-          return when {
+          when {
             targetPath.startsWith('/') -> {
               val fileRoot = project.virtualFileManager.getFsFile(Path.of("/")) ?: return null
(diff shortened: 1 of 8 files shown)


src/main/kotlin/org/pkl/lsp/ast/PklModuleUri.kt
```diff
@@ -202,6 +210,23 @@ class PklModuleUriImpl(project: Project, override val parent: PklNode, override
               )
           }
         }
+
+        "modulepath" -> {
```
HT154: The modulepath isn't globbable, right? This seems off to me.

HT154 (commented):
One issue re: globbing, otherwise LGTM

src/main/kotlin/org/pkl/lsp/ast/PklModuleUri.kt
bioball: Forgot about that! Good catch.

stackoverflow (approved):
Looks good

HT154 (approved):

PR #181, opened 2026-05-12
Fix test flakiness

bioball:
Fixes two issues:

* Use `@TempDir` for new modulepath directories, so that two different tests don't trample on each other.
* Add `settingsManager.update` method which updates its modification count, to ensure that cache keys are invalidated across tests.

### DIFF
--- a/src/main/kotlin/org/pkl/lsp/services/SettingsManager.kt
+++ b/src/main/kotlin/org/pkl/lsp/services/SettingsManager.kt
@@ -33,10 +33,10 @@ import org.pkl.lsp.util.ModificationTracker
 
 @Serializable
 data class WorkspaceSettings(
-  var pklCliPath: Path? = null,
-  var grammarVersion: GrammarVersion? = null,
-  var modulepath: List<Path> = emptyList(),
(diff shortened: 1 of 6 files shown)


src/main/kotlin/org/pkl/lsp/services/SettingsManager.kt
```diff
@@ -55,6 +55,11 @@ class SettingsManager(project: Project) : Component(project), ModificationTracke
     }
   }
 
+  fun update(updater: (WorkspaceSettings) -> WorkspaceSettings) {
```
HT154, line 58: Might be nice to give this lambda a receiver instead, but not a blocker:
```suggestion
  fun update(updater: WorkspaceSettings.() -> WorkspaceSettings) {
```

HT154 (approved):

src/main/kotlin/org/pkl/lsp/services/SettingsManager.kt
bioball: I think I'll keep this as is; it's a matter of `it.copy(...)` versus `copy(...)`; the former feels a little easier to read.

Issue #60, opened 2025-02-13
Publish a native executable


bioball:
The LSP should be published as a native executable, so that users do not need to install Java 22+ on their system.

This is currently blocked by https://github.com/oracle/graal/issues/8113

jclsn:
I would go one step further and seriously question the choice of Java for the LSP.

StayBlue:
The GraalVM issue cited here has been closed.

albuquerq:
Any news here?

dhm116:
Using GraalVM `25.0.2`, I was (seemingly) able to produce a native binary (on MacOS) via the following steps:
1. Generate the `reachability-metadata.json` for a `pkl-lsp-<version>.jar` release:
    ```shell
    mkdir -p META-INF/native-image
    java -agentlib:native-image-agent=config-output-dir=META-INF/native-image -jar pkl-lsp-0.6.0.jar
    ```
    This runs the language server, so I had to `ctrl+c` to exit it.
2. Update the `pkl-lsp-<version>.jar` to include the generated metadata file:
    ```shell
    jar --file pkl-lsp-0.6.0.jar -u META-INF/native-image/reachability-metadata.json
    ```
    You can verify that the file was added to the jar via:
    ```shell
    jar --file ./pkl-lsp-0.6.0.jar -t | grep "META-INF/native-image/reachability-metadata.json"
    ```
3. Build the native image:
    ```shell
    native-image -march=native -jar pkl-lsp-0.6.0.jar
    ```
    You should now have a `pkl-lsp-<version>` native binary in the current directory 👍

mraesener-aubex:
just a quick comment from my side - since Java has (for a reason?) a really bad reputation nowadays, getting rid of the hard dependency having Java installed on your machine (I give Pkl a try especially because of LSP features and tooling) might be a boost in Pkl adoption in general. just my 5 cents and greetings.

udf2457:
Ironically Apple stopped shipping Java on macOS as of OS X 10.7 (Lion), and deprecated it from OS X 10.6 Update 3.

So Mac users have not had a JRE out-of-the-box since 2011.

I therefore agree with @mraesener-aubex , its weird to require people to install a JRE just to use `pkl`.

I absolutely agree that such a prerequisite will adversely affect community adoption.

It is very unfortunate that Apple did not code this in `go` or `rust` which can ship binaries.

colelawrence:
[...]

PR #185, opened 2026-05-18
impl(dev): add native binary release workflow

colelawrence:
My agent one-shotted this. It's based on the TypeScript-Go gh-actions.

Opening as a draft pr in case it's useful reference, since you can try out the build.
https://github.com/colelawrence/pkl-lsp/releases/tag/0.6.0-native.20260518.5

- Build GraalVM native binaries from the release jar for common desktop/server targets.
- Publish versioned tarballs with checksums from a fork-safe generated GitHub Actions workflow.

Related to #60

### DIFF
--- a/.github/index.pkl
+++ b/.github/index.pkl
@@ -240,6 +240,300 @@ release {
   }
 }
 
+workflows {
+  ["workflows/native-release.yml"] {
+    name = "Release Native Binaries"
+    on {
+      push {
+        branches { "main" }
+        `tags-ignore` { "**" }
+        paths {
+          "build.gradle.kts"
+          "buildSrc/**"
+          "gradle/**"
+          "gradle.properties"
+          "settings.gradle.kts"
+          "src/**"
+          ".github/index.pkl"
+        }
+      }
+      workflow_dispatch {}
+    }
+    concurrency {
+      group = "native-release-\(context.github.ref)"
+      `cancel-in-progress` = true
+    }
+    permissions {
+      contents = "read"
+    }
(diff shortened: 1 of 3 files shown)


HT154 (changes requested):
Thanks for the contribution!

We're not able to accept this PR as-is, primarily because it provides no mechanism for building and testing native builds of pkl-lsp locally and does not follow the Pkl project's conventions.

Please see how [apple/pkl](https://github.com/apple/pkl) builds native binaries for `pkl` and `pkldoc`; we would expect those conventions to be followed here for pkl-lsp. In particular:
* Installation of GraalVM and the entire native build should be handled by Gradle
* There should also be variants of the `verifyDistribution` task that target the built native binaries.
* We avoid GHA matrix jobs in favor of using Pkl features like for generators instead.

colelawrence:
> Thanks for the contribution!
> 
> We're not able to accept this PR as-is, primarily because it provides no mechanism for building and testing native builds of pkl-lsp locally and does not follow the Pkl project's conventions.
> 
> Please see how apple/pkl builds native binaries for pkl and pkldoc; we would expect those conventions to be followed here for pkl-lsp. In particular:
> 
> Installation of GraalVM and the entire native build should be handled by Gradle
> There should also be variants of the verifyDistribution task that target the built native binaries.
> We avoid GHA matrix jobs in favor of using Pkl features like for generators instead

@HT154 - not a problem! You can close this. I only intended to open it as reference for the project to see a way for it to be done and distributed.

PR #186, opened 2026-05-19
Fix listChildren in ModulepathResolver

bioball:


build.gradle.kts
```diff
@@ -28,6 +28,7 @@ plugins {
   signing
   zig
   alias(libs.plugins.kotlin)
+  alias(libs.plugins.kotlinPowerAsserts)
```
bioball, line 31: Trying out Kotlin's new power assertions feature!

See https://kotlinlang.org/docs/power-assert.html

HT154 (approved):

PR #188, opened 2026-05-21
Properly flow type `Any.getClass()` calls

HT154:


bioball (approved):
Can you add some test coverage? A hover test should do the trick.

PR #206, opened 2026-06-12
Gracefully handle glob patterns that match large result sets

bioball:
This addresses a perf issue where the language client can slow to a halt for glob patterns that match too many modules.

For these cases:

* Stop expanding a glob once 300 elements has been reached
* Show a diagnostics hint that the glob is too large
* Only provide goto definitions for the first 300
* Simplify type analysis to just `Mapping<String, Module>`

### DIFF
--- a/src/main/kotlin/org/pkl/lsp/ast/Extensions.kt
+++ b/src/main/kotlin/org/pkl/lsp/ast/Extensions.kt
@@ -36,6 +36,7 @@ import org.pkl.lsp.type.Type
 import org.pkl.lsp.type.TypeParameterBindings
 import org.pkl.lsp.type.computeResolvedImportType
 import org.pkl.lsp.type.toType
+import org.pkl.lsp.util.GlobResolver
 import org.pkl.lsp.util.ModificationTracker
 
 val PklClass.supertype: PklType?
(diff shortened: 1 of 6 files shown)


src/main/kotlin/org/pkl/lsp/ast/Extensions.kt
```diff
@@ -349,18 +348,22 @@ class SimpleModuleResolutionResult(val resolved: PklModule?) : ModuleResolutionR
...
+    if (resolved == null) return Type.Unknown
+    if (resolved.exceededMaxElements || resolved.elements.isEmpty())
       return base.mappingType.withTypeArguments(base.stringType, base.moduleType)
     val allTypes =
-      resolved.map {
-        it.computeResolvedImportType(base, bindings, context, preserveUnboundedVars) as Type.Module
+      resolved.elements.map {
+        val module = it.getModule().get()
+        module.computeResolvedImportType(base, bindings, context, preserveUnboundedVars)
+          as Type.Module
```
stackoverflow, line 367: Is this still safe? Before, `resolvedGlob` would filter non-directories and filterNotNull the actual module. Now I don't see any guardrails around this type cast.

src/main/kotlin/org/pkl/lsp/features/GoToDefinitionFeature.kt
```diff
@@ -26,63 +25,100 @@ import org.pkl.lsp.Component
...
+    val resolved = parent.moduleUri?.resolveGlob(context) ?: return NO_COMPLETIONS
+    val futures = buildList {
+      for (file in resolved.elements) {
+        if (!file.isDirectory) {
+          add(file.getModule())
+        }
+      }
     }
+    return futures.sequence().thenApply { elems ->
+      val elems = elems.mapNotNull { it?.locationLink(stringContentsSpan) }
```
stackoverflow: nit: shadowing.

src/main/kotlin/org/pkl/lsp/util/GlobResolver.kt
```diff
@@ -22,13 +22,16 @@ import kotlin.collections.ArrayList
 import org.pkl.lsp.VirtualFile
 
 object GlobResolver {
+  data class GlobResult(val elements: List<VirtualFile>, val exceededMaxElements: Boolean)
 
   private val cache = Collections.synchronizedMap(WeakHashMap<String, Pattern>())
 
   private const val NULL = 0.toChar()
 
+  const val MAX_GLOB_ELEMENTS = 300
```
stackoverflow, line 36: nit: I'd add a doc here like we did for `MAX_LIST_ELEMENTS`.

stackoverflow (approved):

PR #210, opened 2026-06-19
Add native executable builds

### LINKED ISSUE #60: Publish a native executable

pngdeity:
## Summary

This PR adds native executable builds for pkl-lsp using GraalVM native-image. Native executables are built for 6 platform targets: macOS aarch64,
macOS amd64, Linux amd64, Linux aarch64, Alpine Linux amd64, and Windows amd64.

## Conventions

Per [feedback on #185](https://github.com/apple/pkl-lsp/pull/185), the build infrastructure follows the following patterns:

- **Gradle-managed GraalVM installation**: `InstallGraalVm` task downloads, verifies (SHA-256), and installs GraalVM — no `graalvm/setup-graalvm` action
- **`buildSrc` plugin architecture**: `pklGraalVm`,
  `pklJavaExecutable`, `pklNativeExecutable`, `pklNativeLifecycle`
  precompiled script plugins mirror apple/pkl's `build-logic`
  structure
- **`verifyDistribution` variant**: `verifyNativeDistribution` task
  sends LSP messages to the native binary and verifies responses
- **`testStartNativeExecutable` + `checkGlibc`**: matching apple/pkl's
  verification tasks (glibc ≤ 2.17 check, `--version` output
  verification)
- **Pkl-generated CI workflows**: no GHA matrix jobs; native build
  jobs are generated from `.github/index.pkl` using Pkl features
- **UBI8 container** for Linux glibc compatibility,
  `install_musl.sh` for Alpine (from-source musl + zlib with GPG
  verification), `cc_macos_amd64.sh` for macOS amd64
  cross-compilation

## FFM API / Project Panama challenge

pkl-lsp uses
[jtreesitter](https://github.com/tree-sitter/java-tree-sitter) for parsing, which calls 129 tree-sitter C functions via the Java FFM API (`java.lang.foreign.SymbolLookup.libraryLookup()` +
`Linker.downcallHandle()`).

GraalVM 25 supports the FFM API in native-image by default, but requires function descriptors to be registered at build time via`reachability-metadata.json`. If descriptors are missing, the native binary crashes with `IllegalStateException` when the unregistered function is first called.

This PR uses a two-pronged approach:

1. **`--initialize-at-build-time=`** (matching apple/pkl) —
   initializes all classes at build time
2. **Tracing agent with realistic LSP session** — the
   `generateReachabilityMetadata` task runs the LSP with
   `-agentlib:native-image-agent`, sending a realistic session
   (`initialize` → `didOpen` with a sample Pkl file → `completion` →
   `hover` → `didChange` → `didClose` → `shutdown` → `exit`). This
   exercises tree-sitter parsing code paths, capturing FFM downcall
   descriptors in `reachability-metadata.json`.

Note: `--initialize-at-build-time=` alone does not solve the FFM
descriptor registration problem — it controls *when classes are
initialized*, not *which function signatures native-image generates
code for*. Both are needed.

## ExecutableJar re-introduction

PR #31 deliberately removed `ExecutableJar` because "the LSP is not
a user-facing tool." This PR re-introduces it because:

1. Native builds make the LSP a user-facing tool
2. apple/pkl's `pklNativeExecutable` plugin assumes
   `pklJavaExecutable` is applied
3. The `testStartJavaExecutable` task (which verifies the Java
   executable starts correctly) is added to `check`, matching
   apple/pkl's convention

## CI workflows

Native build jobs are triggered by adding `[native]` (or a
platform-specific variant like `[native-macos]`,
`[native-linux-amd64]`) to the PR body. macOS jobs require
`github.repository_owner == 'apple'` (self-hosted runners). Release
builds include native artifacts in the GitHub release.

## `check` behavior change

`testStartJavaExecutable` is now wired into `check` (matching
apple/pkl's convention). This means `./gradlew check` will build the
Java executable and verify it starts correctly. This requires `java`
to be on the PATH.

Closes #60


build.gradle.kts
```diff
@@ -110,6 +112,162 @@ tasks.jar {
...
+      val sampleUri = "${root}tracing-sample.pkl"
+      val samplePkl =
+        """
+        module foo
+
+        import "pkl:base"
+
+        x: Int = 1
+        y: String = "hello"
+        z: List<Int> = [1, 2, 3]
```
bioball:
```suggestion
        z: List<Int> = List(1, 2, 3)
```

build.gradle.kts
```diff
@@ -110,6 +112,162 @@ tasks.jar {
...
+          .replace("\t", "\\t")
+
+      val lspMessages = buildString {
+        fun writeMessage(json: String) {
+          val body = json.trim()
+          append("Content-Length: ${body.toByteArray().size}\r\n\r\n$body")
+        }
+        writeMessage(
+          """{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"processId":null,"rootUri":"$root","workspaceFolders":[{"uri":"$root","name":"pkl-lsp"}],"capabilities":{"workspace":{"workspaceFolders":true,"didChangeConfiguration":{"dynamicRegistration":false}}}}}"""
+        )
```
bioball: Nice, this is great!

It would be great to let lsp4j form these JSON messages, instead of authoring them as plain JSON strings. I can think of two ways to do this:

1. Include `lsp4j` as a buildscript dependency, then create these messages as POJO's provided by LSP4J
2. Create a different source set, where that just calls `PklLsp.run`. Where, that sourceSet extends from the main source set and sends messages into the input stream.

gradle/libs.versions.toml
```diff
@@ -14,18 +14,29 @@ nexusPublishPlugin = "2.0.0"
...
 # Use `../scripts/update_zig.sh <version>` to update these
-zig = "0.16.0" # used for cross-compiling tree-sitter libs
+zig = "0.16.0"                                                                               # used for cross-compiling tree-sitter libs
 zigSha256-macos-aarch64 = "b23d70deaa879b5c2d486ed3316f7eaa53e84acf6fc9cc747de152450d401489"
 zigSha256-linux-amd64 = "70e49664a74374b48b51e6f3fdfbf437f6395d42509050588bd49abe52ba3d00"
 zigSha256-windows-amd64 = "68659eb5f1e4eb1437a722f1dd889c5a322c9954607f5edcf337bc3684a75a7e"
 zigSha256-linux-aarch64 = "ea4b09bfb22ec6f6c6ceac57ab63efb6b46e17ab08d21f69f3a48b38e1534f17"
 
+# GraalVM for native-image builds
+# 25.0.2 no longer supports macos-x64, so we pin to 25.0.1
```
bioball, line 30: We can go ahead and use 25.0.2 here; Pkl itself is dropping macOS x64 soon enough too.

bioball (commented):
Thanks, this is great! Some initial comments.

PR #211, opened 2026-06-23
Add analysis for extends clause

bioball:
Also, this adds a quickfix for implementing members.

* Insert members and also imports when needed
* Refactor: merge `shortDisplayName` and `name` (they are duplicate properties that mean the same thing)
* Move `ModifierQuickfixTest` to `PklAddModifierQuickFixTest`; make sibling of new test
* Fix add modifier logic when no module header exists

### DIFF
--- a/src/main/kotlin/org/pkl/lsp/analyzers/ExtendsClauseAnalyzer.kt
+++ b/src/main/kotlin/org/pkl/lsp/analyzers/ExtendsClauseAnalyzer.kt
@@ -36,11 +36,11 @@ import org.pkl.lsp.ast.TokenType
 import org.pkl.lsp.ast.declaredMethods
 import org.pkl.lsp.ast.declaredProperties
 import org.pkl.lsp.ast.effectiveParentProperties
+import org.pkl.lsp.ast.hasDefault
 import org.pkl.lsp.ast.isInStdlib
 import org.pkl.lsp.ast.methods
 import org.pkl.lsp.ast.resolve
 import org.pkl.lsp.packages.dto.PklProject
-import org.pkl.lsp.type.toType
 
 class ExtendsClauseAnalyzer(project: Project) : Analyzer(project) {
   override fun doAnalyze(node: PklNode, diagnosticsHolder: DiagnosticsHolder): Boolean {
@@ -189,8 +189,7 @@ class ExtendsClauseAnalyzer(project: Project) : Analyzer(project) {
       if (member !is PklClassProperty || !member.isFixedOrConst) continue
       val definedProperty = def.declaredProperties.find { it.name == member.name }
       if (definedProperty != null) continue
-      if (member.expr != null) continue
-      if (member.type.toType(base, mapOf(), context).hasDefault(base, context)) continue
+      if (member.hasDefault(base, context)) continue
 
       // copy Java/Kotlin error message and provide information about just the first missing
       // property.
--- a/src/main/kotlin/org/pkl/lsp/analyzers/ExtendsClauseAnalyzer.kt
+++ b/src/main/kotlin/org/pkl/lsp/analyzers/ExtendsClauseAnalyzer.kt
@@ -1,0 +1,277 @@
+/*
+ * Copyright © 2026 Apple Inc. and the Pkl project authors. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.pkl.lsp.analyzers
+
+import org.pkl.lsp.ErrorMessages
+import org.pkl.lsp.PklBaseModule
+import org.pkl.lsp.Project
+import org.pkl.lsp.actions.PklAddModifierQuickFix
+import org.pkl.lsp.actions.PklImplementMembersQuickFix
+import org.pkl.lsp.ast.PklClass
+import org.pkl.lsp.ast.PklClassExtendsClause
+import org.pkl.lsp.ast.PklClassMember
+import org.pkl.lsp.ast.PklClassProperty
+import org.pkl.lsp.ast.PklDeclaredType
+import org.pkl.lsp.ast.PklModule
+import org.pkl.lsp.ast.PklModuleExtendsAmendsClause
+import org.pkl.lsp.ast.PklModuleType
+import org.pkl.lsp.ast.PklNode
+import org.pkl.lsp.ast.PklType
+import org.pkl.lsp.ast.PklTypeAlias
+import org.pkl.lsp.ast.PklTypeDefOrModule
+import org.pkl.lsp.ast.TokenType
+import org.pkl.lsp.ast.declaredMethods
+import org.pkl.lsp.ast.declaredProperties
+import org.pkl.lsp.ast.effectiveParentProperties
+import org.pkl.lsp.ast.isInStdlib
+import org.pkl.lsp.ast.methods
+import org.pkl.lsp.ast.resolve
+import org.pkl.lsp.packages.dto.PklProject
+import org.pkl.lsp.type.toType
+
+class ExtendsClauseAnalyzer(project: Project) : Analyzer(project) {
+  override fun doAnalyze(node: PklNode, diagnosticsHolder: DiagnosticsHolder): Boolean {
+    when (node) {
+      is PklModuleExtendsAmendsClause -> checkModuleExtendsClause(node, diagnosticsHolder)
+      is PklClassExtendsClause -> checkClassExtendsClause(node, diagnosticsHolder, node.type)
+    }
+    return true
+  }
+
+  private fun checkModuleExtendsClause(
+    clause: PklModuleExtendsAmendsClause,
+    holder: DiagnosticsHolder,
+  ) {
+    if (!clause.isExtend) return
+    val moduleUri = clause.moduleUri ?: return
+    val context = clause.enclosingModule?.virtualFile?.pklProject
+    val resolved =
+      moduleUri.resolve(context) ?: return // checked by [PklModuleUriAndVersionAnnotator]
+    if (!resolved.isAbstractOrOpen) {
+      val moduleName = resolved.name
+      holder.addError(
+        moduleUri.stringConstant,
+        ErrorMessages.create("cannotExtendModule", moduleName),
+      ) {
+        if (resolved.virtualFile.canModify()) {
+          actions +=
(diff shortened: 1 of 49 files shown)


src/test/kotlin/org/pkl/lsp/actions/PklImplementMembersQuickFixTest.kt
```diff
@@ -0,0 +1,465 @@
...
+          abstract fixed name: String
+        }
+        class Child extends Base {
+          fixed name: String = TODO()
+        }
+        """
+          .trimIndent(),
+    )
+  }
+
```
stackoverflow, line 161: The quickfix is adding the property out of the class when there's already something in the body:
```suggestion

  @Test
  fun `quickfix implements member in class with existing body`() {
    checkQuickFix(
      before =
        """
        abstract module Foo

        abstract class Base {
          abstract name: String
        }
        class Child extends Base {
          bar: Int = 5
        }
        """
          .trimIndent(),
      after =
        """
        abstract module Foo

        abstract class Base {
          abstract name: String
        }
        class Child extends Base {
          bar: Int = 5

          name: String = TODO()
        }
        """
          .trimIndent(),
    )
  }
```

src/test/kotlin/org/pkl/lsp/actions/PklImplementMembersQuickFixTest.kt
```diff
@@ -0,0 +1,465 @@
...
+  @Test
```
stackoverflow, line 162: The quickfix is fixing properties that already have a default.
```suggestion
  @Test
  fun `quickfix does not reimplement fixed property that has an inherited default`() {
    checkQuickFix(
      before =
        """
        abstract module Foo

        abstract class Base {
          fixed name: String = "default"
          abstract function greet(): String
        }
        class Child extends Base {}
        """
          .trimIndent(),
      after =
        """
        abstract module Foo

        abstract class Base {
          fixed name: String = "default"
          abstract function greet(): String
        }
        class Child extends Base {
          function greet(): String = TODO()
        }
        """
          .trimIndent(),
    )
  }
  
  @Test
```

src/test/kotlin/org/pkl/lsp/GoToDefinitionPackagesTest.kt
```diff
@@ -60,27 +60,27 @@ class GoToDefinitionPackagesTest : LspTestBase() {
...
       }
-    }
-    """
+      """
         .trimIndent()
     )
     val resolved = goToDefinition().single()
     assertThat(resolved).isInstanceOf(PklClassProperty::class.java)
-    assertThat(resolved.enclosingModule!!.moduleName).isEqualTo("k8s.api.apps.v1.Deployment")
+    assertThat(resolved.enclosingModule!!.name).isEqualTo("k8s.api.apps.v1.Deployment")
```
stackoverflow: This assertion is failing. `name` only returns the last part, not the fully qualified name.

src/test/kotlin/org/pkl/lsp/actions/PklImplementMembersQuickFixTest.kt
```diff
@@ -0,0 +1,465 @@
...
+      import "other.pkl" as bar
+
+      class Child extends foo.Base {
+        bar: bar2.Qux = TODO()
+      }
+      """
+        .trimIndent()
+    )
+  }
+
```
stackoverflow, line 422: This is failing. Pretty sure is due to the auto-importer making assumptions about the order of imports.
```suggestion

 @Test
  fun `import type in different module, with chained alias conflicts`() {
    createPklFile(
      "foo.pkl",
      """
      abstract module foo

      import "bar.pkl"

      abstract class Base {
        abstract bar: bar.Qux
      }
      """
        .trimIndent(),
    )

    createPklFile(
      "bar.pkl",
      """
      class Qux
      """
        .trimIndent(),
    )

    createPklFile(
      "test.pkl",
      """
      import "aaa.pkl" as bar2
      import "foo.pkl"
      import "other.pkl" as bar

      class Child extends foo.Base
      """
        .trimIndent(),
    )
    implementMembersAndAssertResultEqualTo(
      """
      import "aaa.pkl" as bar2
      import "bar.pkl" as bar3
      import "foo.pkl"
      import "other.pkl" as bar

      class Child extends foo.Base {
        bar: bar3.Qux = TODO()
      }
      """
        .trimIndent()
    )
  }
  
```

src/test/kotlin/org/pkl/lsp/actions/PklImplementMembersQuickFixTest.kt
```diff
@@ -0,0 +1,465 @@
...
+    implementMembersAndAssertResultEqualTo(
+      """
+      extends "foo.pkl"
+
+      bar: Int = TODO()
+      """
+        .trimIndent()
+    )
+  }
+
```
stackoverflow, line 260: `lastMember` might return `null` if the module has no imports and properties/methods.
```suggestion

  @Test
  fun `quickfix implements module members without deleting trailing members`() {
    createPklFile(
      "foo.pkl",
      """
      abstract module foo

      abstract bar: Int
      """
        .trimIndent(),
    )

    createPklFile(
      "test.pkl",
      """
      extends "foo.pkl"

      class Bar {}
      """
        .trimIndent(),
    )
    implementMembersAndAssertResultEqualTo(
      """
      extends "foo.pkl"

      bar: Int = TODO()

      class Bar {}
      """
        .trimIndent()
    )
  }
  
```

src/main/kotlin/org/pkl/lsp/ast/Extensions.kt
```diff
@@ -659,3 +661,184 @@ fun Appendable.renderParameterList(
...
+    }
+  }
+  return false
+}
+
+private fun appendNumericSuffix(name: String): String =
+  when (name.last()) {
+    in '0'..'9' -> {
+      val number = name.takeLastWhile { it in '0'..'9' }
+      name.substring(0, name.length - number.length) + (number.toInt() + 1)
```
stackoverflow, line 737: `toInt()` might overflow if the suffix is bigger than MAX_INT. Better to use a BigInt.
Yeah, it's a rare edge case, but we can't make assumptions about people's code.

stackoverflow (changes requested):

src/test/kotlin/org/pkl/lsp/actions/PklImplementMembersQuickFixTest.kt
bioball: Great catch!

src/test/kotlin/org/pkl/lsp/actions/PklImplementMembersQuickFixTest.kt
bioball: Great catch!

src/test/kotlin/org/pkl/lsp/actions/PklImplementMembersQuickFixTest.kt
bioball: Nice find!

src/main/kotlin/org/pkl/lsp/ast/Extensions.kt
bioball: This doesn't strike me as very important to handle. Likely, more things will blow up if you have that many imports. For example, our `Span` has `beginLine: Int`, `beginCol: Int`. Also, Pkl itself will probably blow up if trying to eval a module like this.

src/test/kotlin/org/pkl/lsp/actions/PklImplementMembersQuickFixTest.kt
bioball: Good catch!

src/test/kotlin/org/pkl/lsp/GoToDefinitionPackagesTest.kt
bioball: Ah, I re-introduced `moduleName`.

src/main/kotlin/org/pkl/lsp/ast/Extensions.kt
stackoverflow: I was thinking more about having a single (or couple) imports with huge numbers as suffix, not that many imports. But I guess in that case this code would just pick a small number, right?

bioball: Yeah; just starts adding numeric suffixes until there's no conflict.

src/test/kotlin/org/pkl/lsp/actions/PklImplementMembersQuickFixTest.kt
```diff
@@ -0,0 +1,575 @@
...
+        class Child extends Base2 {
+          name: String = TODO()
+
+          function greet(): String = TODO()
+        }
+        """
+          .trimIndent(),
+    )
+  }
+
```
stackoverflow, line 231: This is failing (`name` is being re-implemented). I think this is due to `ExtendsClauseAnalyzer.checkAbstractMembers` and `PklImplementMembersQuickFix.getEdits` having different rules to wheter/which members get inserted.
```suggestion

@Test
  fun `quickfix only implements members not already declared`() {
    checkQuickFix(
      before =
        """
        abstract module Foo

        abstract class Base {
          abstract name: String
          abstract age: Int
        }
        class Child extends Base {
          name: String = "Bob"
        }
        """
          .trimIndent(),
      after =
        """
        abstract module Foo

        abstract class Base {
          abstract name: String
          abstract age: Int
        }
        class Child extends Base {
          name: String = "Bob"

          age: Int = TODO()
        }
        """
          .trimIndent(),
    )
  }
```

src/main/kotlin/org/pkl/lsp/analyzers/ExtendsClauseAnalyzer.kt
```diff
@@ -0,0 +1,276 @@
...
+            PklAddModifierQuickFix(
+              resolved,
+              ErrorMessages.create("addModuleModifier", moduleName, "abstract"),
+              TokenType.ABSTRACT,
+            )
+        }
+      }
+    }
+    val module = clause.enclosingModule ?: return
+    checkParentClassDef(clause, module, project.pklBaseModule, holder, context)
```
stackoverflow, line 86: Not entirely sure, but this might generate a stack overflow for a circular inheritance (`effectiveParentProperties` -> `parentTypeDef` -> `cache`). Might be a good idea to create a test just to be sure.

stackoverflow (approved):
One issue left. Approving to unblock.

src/test/kotlin/org/pkl/lsp/actions/PklImplementMembersQuickFixTest.kt
bioball: Ah, whoops, that was a bug I introduced in my latest commit. Fixed!

src/main/kotlin/org/pkl/lsp/analyzers/ExtendsClauseAnalyzer.kt
bioball: This does stackoverflow! But I think this is an orthogonal bug in our resolving logic; merging this for now and we can fix in a follow-up.

PR #224, opened 2026-07-07
Enforce stricter rules on reference access

HT154:
Matches behavior changes from https://github.com/apple/pkl/pull/1718


stackoverflow (approved):
LGTM. Same nit as the intellij PR.

PR #230, opened 2026-07-09
Improve handling of doc comment reference links

### LINKED ISSUE #47: Incorrect formatting of code blocks that look like member links
When processing member links in doc comments, things inside code blocks _also_ get formatted, which is incorrect.

This doc comment:

````
This is foo
```
bar = (foo) { [baz] = 1 }
```
````

Gets turned into:

````
This is foo
```
bar = (foo) { `baz` = 1 }
```
````

To fix this, we maybe should introduce a markdown parser.

### PR DESCRIPTION
**Pull request by bioball**
This improves the handling of member links in doc comments!

* Syntax highlighting
* Hover shows resolved node documentation
* Go-to-definition goes to resolved node
* Diagnostics for unresolved links
* Discover member links using markdown parser instead of regex
* Mark local as used if referenced in member link

This also fixes an issue where we incorrectly process snippets within code blocks as member links.

Closes https://github.com/apple/pkl-lsp/issues/47

### CODE REVIEW

File: src/main/kotlin/org/pkl/lsp/ast/Basic.kt [hunk 1]
```diff
@@ -67,3 +76,164 @@ class PklShebangCommentImpl(project: Project, parent: PklNode, override val ctx:
...
+    }
+
+  override val processedContents: String
+    get() {
+      var myContents = contents
+      for (link in
+        memberLinks.sortedWith(
+          Comparator.comparing<PklDocCommentMemberLink, Int> { -it.span.beginCol }
+            .thenComparing { -it.span.beginLine }
+        )) {
```
HT154, line 103: Took me a minute to understand why this sort happens: it's so that links are replaced end-to-start so that fiddly replacement offsets don't need to be tracked as you go. Might be good to comment this since it's not immediately obvious.

src/main/kotlin/org/pkl/lsp/ast/PklNode.kt
```diff
@@ -187,30 +189,18 @@ interface PklModifierListOwner : PklNode {
...
-    get() {
-      val doc = docComment?.text?.trim() ?: return null
-      return doc.lines().joinToString("\n") { it.trimStart().removePrefix("///") }.trimIndent()
-    }
-
-  val parsedComment: String?
-    get() {
-      return rawComment?.let { DocCommentMemberLinkProcessor.process(it, this) }
-    }
+  val myDocumentation: String?
```
HT154: Nit: naming-wise, `myX` feels like it should be a private field. Maybe `docContents`  as an alternative?

src/main/kotlin/org/pkl/lsp/LspUtil.kt
```diff
@@ -329,3 +329,8 @@ fun ensureJarFileSystem(uri: URI) {
 fun getDoc(node: PklDocCommentOwner, context: PklProject?): Either<String, MarkupContent> {
   return Either.forRight(MarkupContent("markdown", node.effectiveDocComment(context) ?: ""))
 }
+
+val memberLinkKeywords = setOf("null", "true", "false", "this", "unknown", "nothing")
```
HT154, line 333: Should `module` be included here as well?

src/main/kotlin/org/pkl/lsp/resolvers/ResolveVisitors.kt
```diff
@@ -493,16 +496,25 @@ object ResolveVisitors {
...
+        if (forMemberLink) {
+          item.insertText = "$name()"
+        } else {
+          item.insertText = "$name($strPars)"
+        }
+        if (forMemberLink) {
+          item.insertTextFormat = InsertTextFormat.Snippet
+        } else {
+          item.insertTextFormat = InsertTextFormat.PlainText
+        }
```
HT154, line 514: Any reason not to combine these?
```suggestion
        if (forMemberLink) {
          item.insertText = "$name()"
          item.insertTextFormat = InsertTextFormat.Snippet
        } else {
          item.insertText = "$name($strPars)"
          item.insertTextFormat = InsertTextFormat.PlainText
        }
```

HT154 (commented):
Awesome!! Just a few small things. Approving to unblock.

src/main/kotlin/org/pkl/lsp/ast/Basic.kt
```diff
@@ -67,3 +76,164 @@ class PklShebangCommentImpl(project: Project, parent: PklNode, override val ctx:
...
+        val replacement =
+          when (
+            val resolved = DocCommentResolvers.resolveLink(project, link.text, parent!!, true)
+          ) {
+            is PklNode -> {
+              buildString {
+                append("[`${link.text}`](")
+                append(resolved.getLocationUri(forDocs = true))
+                append(")")
+              }
```
HT154, line 113: Looks like this might not be properly handling labeled member links `[<label>][<reference>]`

src/main/kotlin/org/pkl/lsp/LspUtil.kt
bioball: Ah, yes! It's also missing in IntelliJ.

src/main/kotlin/org/pkl/lsp/ast/Basic.kt
bioball: Ah, good catch! Fixed

HT154 (approved):

PR #243, opened 2026-09-01
Remove wrong scopeUri

stackoverflow:
`scopeUri` is supposed to be an actual URI that scopes this request to a specific folder/workspace. As `Pkl` is not a valid URI, vscode interprets that as `file:///Pkl` which doesn't match any workspace, defaulting to the global configuration (which is what we want). For other editors, they just ignore `scopeUri`. So, this change has no visible effect.
We can decide in the future to scope configuration requests to specific workspaces.


bioball (approved):
LGTM

HT154:
Does this break or otherwise affect users' existing saved settings?

stackoverflow:
> Does this break or otherwise affect users' existing saved settings?

No. Most editors (like neovim) ignore `scopeUri`.
Even in vscode, you'd need to have your files inside `/Pkl` **and** have workspace specific settings for these configs there, overwriting the global ones.

HT154 (approved):

